PR #55, opened 2022-11-08
Fix seg fault

caleb-johnson:
CC `reconstruct_full_distribution` was causing seg faults in our demos.

This PR attempts to more appropriately allocate threads for multi-core post-processing.


garrison (approved):
If this fixes the issue, then sure, let's merge it.  But I am concerned that we may not be able to get much benefit from the thread pool.  Specifically, `naive_compute` is CPU bound, and with Python's global interpreter lock, probably only one thread is able to execute at a time. Some of the `numpy` calls within that function might be able to release the GIL, and if so, we might gain some advantage from the threads, but it's not a priori obvious.  To know whether the threads provide any advantage, we'd have to benchmark.

garrison:
Also, was any error message shown before the segfault, or no?  And does the segfault happen intermittently, or all the time?  I haven't witnessed it myself, either locally or on CI.

garrison (changes requested):
Actually, retracting my earlier approval now that I noticed the breaking tests ;)

caleb-johnson:
> Actually, retracting my earlier approval now that I noticed the breaking tests ;)

This notebook isn't hanging locally. Going to have to take a closer look at what's going on here in the CI

PR #58, opened 2022-11-11
Remove Qiskit Runtime retry loop

garrison:
This removes the retry loop which would handle when Qiskit Runtime would return a result of `None`.  The bug that caused this issue has been fixed in the service, so the retry loop is no longer necessary.

I'm in _no hurry_ to merge this, but I wanted to open a PR nonetheless before we forget about it.


caleb-johnson:
I support this whenever you think it's ready. I believe this call should be much more stable now

PR #64, opened 2022-12-06
WIP: Add option to fix value associated with the HF bitstring

caleb-johnson:


garrison (commented):
This looks like a correct and fine addition.  I believe my only comment is that it's not clear to me what it means precisely to "fix" the first HF value (the code/comment implies it is dropped from the tensor prep circuits?), and why one might want to do this.  I wonder if there's an improvement that could be made to another section of the documentation other than the docstrings (e.g., explanatory material).

caleb-johnson:
> This looks like a correct and fine addition. I believe my only comment is that it's not clear to me what it means precisely to "fix" the first HF value (the code/comment implies it is dropped from the tensor prep circuits?), and why one might want to do this. I wonder if there's an improvement that could be made to another section of the documentation other than the docstrings (e.g., explanatory material).

Yes, there is some good documentation on this feature in the prototype. I will move it over in this PR

PR #70, opened 2023-03-28
Speed up kronecker product

hitomitak:
Rewrote Kronecker product calculation to not use `np.kron` function. This is because using this numpy function for vector calculations is very computationally expensive.  This PR can speed up the reconstruction time as the following image. 

Experiment Environment:
- Cut random circuits of 29 and 30 qubit
- max_subcircuit_width=10, max_cut=10

image

### DIFF
--- a/circuit_knitting_toolbox/circuit_cutting/wire_cutting/wire_cutting_post_processing.py
+++ b/circuit_knitting_toolbox/circuit_cutting/wire_cutting/wire_cutting_post_processing.py
@@ -371,9 +371,11 @@ def naive_compute(
             if summation_term_prob is None:
                 summation_term_prob = subcircuit_entry_prob
             else:
-                summation_term_prob = np.kron(
-                    summation_term_prob, subcircuit_entry_prob
-                )
+                summation_term_prob = (
+                    summation_term_prob[:, None,]*subcircuit_entry_prob[None, :,]
+                    ).reshape(
+                        summation_term_prob.size * subcircuit_entry_prob.size
+                    ) 
                 overhead["multiplications"] += len(summation_term_prob)
         if reconstructed_prob is None:
             reconstructed_prob = summation_term_prob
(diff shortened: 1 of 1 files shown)


garrison:
Thanks for this.  To be honest, it's kind of disappointing that numpy's `kron` is not at least as efficient as the code you give when it is passed a 1D array.

garrison:
I was able to reproduce this in a simple test case using numpy 1.22.2:

```python
>>> import timeit
>>> import numpy as np
>>> a = np.random.rand(1000)
>>> b = np.random.rand(1000)
>>> timeit.timeit(lambda: np.kron(a, b), number=200)
1.302669789060019
>>> timeit.timeit(lambda: (a[:,None] * b[None,:]).reshape(a.size * b.size), number=200)
0.25691784801892936
>>> np.allclose(np.kron(a, b), (a[:,None] * b[None,:]).reshape(a.size * b.size))
True
>>> np.__version__
'1.22.2'
```

However, if I try on numpy 1.24.2, the performance gap closes, and the current version might actually be a bit faster:

```python
>>> import timeit
>>> import numpy as np
>>> a = np.random.rand(1000)
>>> b = np.random.rand(1000)
>>> timeit.timeit(lambda: np.kron(a, b), number=200)
0.2618443979881704
>>> timeit.timeit(lambda: (a[:,None] * b[None,:]).reshape(a.size * b.size), number=200)
0.30681791296228766
>>> np.allclose(np.kron(a, b), (a[:,None] * b[None,:]).reshape(a.size * b.size))
True
>>> np.__version__
'1.24.2'
```

Have you tried upgrading numpy?

garrison:
One alternative idea is to warn in this function if somebody is using an old numpy that is known to be slow.

garrison:
My best guess is that the performance improvement is due to https://github.com/numpy/numpy/pull/21354.

It might be better here to just require the first numpy release that includes that commit as the minimum supported by the toolbox (rather than emitting a warning).

garrison:
I just tried on numpy 1.23.0, and the performance is essentially the same between those two methods.  I would support modifying `pyproject.toml` to require `numpy>=1.23.0`.

PR #73, opened 2023-03-30
Move `MIPModel` import to within a function

garrison:
~~Currently, the code blows up if one imports any wire cutting functionality and cplex is not installed.~~  With this change, users should be able to perform manual wire cutting, even on platforms where cplex is not available.

EDIT: I realized my premise might actually have been wrong.  `mip_model.py` imports from DOcplex, but that is currently available even when cplex is not.

### DIFF
--- a/circuit_knitting_toolbox/circuit_cutting/wire_cutting/wire_cutting.py
+++ b/circuit_knitting_toolbox/circuit_cutting/wire_cutting/wire_cutting.py
@@ -24,7 +24,6 @@ from qiskit_ibm_runtime import Options, QiskitRuntimeService
 from .wire_cutting_evaluation import run_subcircuit_instances
 from .wire_cutting_post_processing import generate_summation_terms, build
 from .wire_cutting_verification import generate_reconstructed_output
-from .mip_model import MIPModel
 
 
 def cut_circuit_wires(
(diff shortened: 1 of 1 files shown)


caleb-johnson:
> ~Currently, the code blows up if one imports any wire cutting functionality and cplex is not installed.~ With this change, users should be able to perform manual wire cutting, even on platforms where cplex is not available.
> 
> EDIT: I realized my premise might actually have been wrong. `mip_model.py` imports from DOcplex, but that is currently available even when cplex is not.

So what exactly issue does this PR resolve, given that mip_model relies only on open software?

I believe DOCPLEX will work as long as the problem is small enough, if the problem is too large for free version, it will look for a CPLEX paid install and use it if possible, otherwise it errors and tells you to pay for it. Is that right?

garrison:
My understanding is that DOcplex always requires cplex to perform solving, but that there are two different versions of cplex: the [community edition](https://pypi.org/project/cplex/) and the [premium](https://www.ibm.com/products/ilog-cplex-optimization-studio/cplex-optimizer) one.  Both are proprietary and only built for a [limited set of architectures](https://pypi.org/project/cplex/#files) (e.g., no M1 Mac) and Python versions (e.g., no 3.11).

garrison:
Closing, as this is (i) included in #109; and (ii) not immediately useful outside of that context.

PR #81, opened 2023-04-05
Fix deprecated usages of Bit

caleb-johnson:
Wire cutting has several deprecated usages of `Bit`, which should be updated.


garrison (approved):
Awesome, I am so glad you did this.  It has been bugging me forever! :100:

PR #82, opened 2023-04-05
A few small improvements

garrison:
I've had this branch sitting locally for a while; I think it's time to contribute it.

- Use appropriate exceptions instead of `AttributeError`
- Use tuple unpacking rather than indexing (will catch errors better)
- raise `NotImplementedError` for methods not implemented
- Remove double variable assignment in `docs/conf.py`


caleb-johnson (approved):
Nice, thanks

PR #84, opened 2023-04-06
Add support for reno release notes

garrison:
This adds support for reno, so that we can have an explicit "release notes" section in our docs.  This also puts us in line with the practices of Qiskit and the Qiskit application modules.

This will be distinct from the PR summary that gets added to the [release page](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/releases) for each release.  It will allow us to more thoroughly document new features, deprecations, and other changes in the toolbox.


caleb-johnson (approved):
Great, thanks

PR #92, opened 2023-04-08
Update usage of Qiskit Estimator

caleb-johnson:
The Estimator no longer needs all circuits up front. Instead, the preferable usage is to just pass in the circuits you want to evaluate when `run` is called.

This PR cleans up some deprecation warnings being thrown during testing


garrison (approved):
Great!

PR #93, opened 2023-04-09
Remove nptyping as dependency

### LINKED ISSUE #91: Remove nptyping as a dependency
This package is noisy with warnings and is no longer worth it. We can use `np.ndarray` as the type for all of these instances

caleb-johnson:
Fixes #91 

nptyping is causing warnings in CI and it is no longer worth it to carry as a dependency.

- Most usages of nptyping were using the flat `NDArray` type, which is 1:1 replaceable with `np.ndarray`
- The usages of nptyping that were more informative were associated with copious comments describing the meaning and shape of those tensors (mostly in the cholesky and knitter modules of EF), so it is reasonable to defer to those comments in favor of more palatable type hints
- `np.typing` module contains more powerful type hinting control if we find a need in the future
- flake8 was flagging some of our previous type hints as having incorrect format, so this eliminates the confusion associated with the cumbersome nptyping type hints (i.e. using `Shape` and `*` correctly)


garrison (approved):
Thanks, this looks very reasonable to me.

PR #95, opened 2023-04-09
Migrate to modern style of type hinting

### LINKED ISSUE #88: Remove deprecated types
The typing package has deprecated `List` and `Tuple` in favor of the base types. Ensure no deprecated types are being used.

caleb-johnson:
Fixes #88 

This PR leverages the more flexible type hinting available in python 3.10+ and updates the CI linter to run checks in python 3.10.

There are still a couple of places in the code where `Union` and `Dict` are used to specify types at runtime, such as when specifying an obnoxiously long type variable or using the `typing.cast` function, as the new type hinting flexibility is achieved by simply opting out of checking annotations. In other words `dict`, `tuple`, `list` etc. are not replacements types, but rather more palatable ways of specifying type hints in source code. More [here](https://bugs.python.org/issue45117)


circuit_knitting_toolbox/circuit_cutting/wire_cutting/wire_cutting_evaluation.py
```diff
@@ -39,14 +42,14 @@ def run_subcircuit_instances(
 
     Args:
         - subcircuits (Sequence[QuantumCircuit]): the list of subcircuits to execute
-        - subcircuit_instances (Dict): dictionary containing information about each of the
+        - subcircuit_instances (dict): dictionary containing information about each of the
             subcircuit instances
         - service (QiskitRuntimeService): the runtime service
         - backend_names (Sequence[str]): the backend(s) used to execute the subcircuits
         - options (Sequence[Options]): options for the runtime execution of subcircuits
```
garrison, line 49: another place where the `None`s could be called out.

On the other hand, this could also serve as an example of why providing explicit type annotations in the docstrings might be something to avoid.  It's just so easy for them to fall out of sync with reality.  Something to keep in mind for the future.

circuit_knitting_toolbox/circuit_cutting/wire_cutting/wire_cutting_post_processing.py
```diff
@@ -69,9 +72,9 @@ def get_label(label_idx: int, num_cuts: int) -> Sequence[str]:
 
 def attribute_label(
     label: Sequence[str],
-    O_rho_pairs: List[Tuple[Dict[str, Union[int, Any]], Dict[str, Union[int, Any]]]],
+    O_rho_pairs: list[tuple[dict[str, int | Any], dict[str, int | Any]]],
```
garrison, line 75: Another `int | Any`.  (I'm not going to comment on any more beyond this one.)

garrison (commented):
Wow, thank you for taking on this task.  Even reviewing it took a lot of effort.

garrison (approved):

PR #109, opened 2023-04-11
Make cplex truly optional

### LINKED ISSUE #90: Make cplex and DOcplex dependencies be truly optional
- [ ] Make it so circuit cutting can be imported without cplex or DOCplex present: #73
- [ ] Modify `pyproject.toml` to make cplex and DOcplex optional dependencies
- [ ] Update tests to skip certain ones that require cplex if it is not present.  (This might not be easily possible for the notebook tests, but we can try.)
- [ ] Update install instructions to encourage installing cplex if automatic cut finding is desired; also, users who don't need it may not need to use Docker.
- [ ] Attempt tests under Python 3.11; add it to test matrix.
- [ ] Add a release note

garrison:
Closes #90

Action items:
- [x] Make it so circuit cutting can be imported without cplex or DOcplex present: #73
- [x] Modify `pyproject.toml` to make cplex and DOcplex optional dependencies
- [x] Update tests to skip certain ones that require cplex if it is not present.
- [x] Ditto for the notebook tests (at least, handle them as best as we can)
- [x] Update install instructions to encourage installing cplex if automatic cut finding is desired; also, users who don't need it may not need to use Docker.
- [x] Attempt tests under Python 3.11; add it to test matrix.
- [x] Add a release note
- [x] Informative error if automatic cut finding fails because DOcplex is not present (cplex is covered by #127)


caleb-johnson:
What will we do with functions which require CPLEX, like [this](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/6f3c8572d7770bfef7530ec73c836b09e7ef741c/circuit_knitting_toolbox/circuit_cutting/wire_cutting/wire_cutting.py#L30)?
Do we just throw an error if they pass `method='automatic'`, and cplex is not installed?


What about functions that will try to load an MIPModel that live under the function above? Will they operate as normal and rely on the MIPModel wrapper function mentioned in #73  to throw the error?

caleb-johnson:
This is great, thanks. Do we want to add something to [find_wire_cuts](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/5d4a89110d695aed16779eecbd5482a001a81c00/circuit_knitting_toolbox/circuit_cutting/wire_cutting/wire_cutting.py#L311) that fails gracefully if cplex isn't installed? Or would it need to be handled when the module is imported, since MIPModel (and thus docplex) are imported when the modules are imported?

garrison:
> I notice https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/73 never made it in

#73 is included in this current changeset, actually.

caleb-johnson (approved):
Really clean implementation here, thanks

Now, users need to pip install circuit-knitting-toolbox[cplex] to get the full range of tools?

garrison:
> Now, users need to pip install circuit-knitting-toolbox[cplex] to get the full range of tools?

Yes.  But this actually won't give pyscf.  I'll make a new PR to add that as an option, too.

PR #110, opened 2023-04-11
Enable ruff linter

garrison:


garrison:
@caleb-johnson, the remaining problem that ruff discovers is:

```
circuit_knitting_toolbox/entanglement_forging/cholesky_decomposition.py:332:9: F841 [*] Local variable `num_gammas` is assigned to but never used
Found 1 error.
```

This is from

https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/389b8b8039b6726ec86996f518c6d8b4cbd82ebb/circuit_knitting_toolbox/entanglement_forging/cholesky_decomposition.py#L325-L333

Should `num_gammas` be used?  If not, we could assign it to `_`, I suppose, to get rid of the error.

caleb-johnson:
> @caleb-johnson, the remaining problem that ruff discovers is:
> 
> ```
> circuit_knitting_toolbox/entanglement_forging/cholesky_decomposition.py:332:9: F841 [*] Local variable `num_gammas` is assigned to but never used
> Found 1 error.
> ```
> 
> This is from
> 
> https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/389b8b8039b6726ec86996f518c6d8b4cbd82ebb/circuit_knitting_toolbox/entanglement_forging/cholesky_decomposition.py#L325-L333
> 
> Should `num_gammas` be used? If not, we could assign it to `_`, I suppose, to get rid of the error.

Hey thanks for this PR. I've been using ruff locally too, and all the output has been giving me anxiety hah!

I think assigning it to _ is right. I've seen a few other unused variables in the ruff output as well in my branch. Not sure which of them were unique to my branch and which came from main

pyproject.toml
```diff
@@ -82,3 +83,14 @@ remove-all-unused-imports = true
 
 [tool.flit.module]
 name = "circuit_knitting_toolbox"
+
+[tool.ruff]
```
caleb-johnson, line 87: Forgive my ignorance, so ruff operates through the build metafile rather than its own ini file?

caleb-johnson (approved):
Awesome, thanks. One question for my own edification

pyproject.toml
garrison: It [looks like](https://github.com/charliermarsh/ruff#configuration) we have the option of using `ruff.toml` or `.ruff.toml` if we so choose, instead of `pyproject.toml`.  I kind of like having all the configuration in `pyproject.toml`, but I think Pedro holds the opposite opinion, and I believe he holds it pretty strongly.  In his projects, typically only `black` is configured in `pyproject.toml`, as it doesn't support any other configuration file.  (I think he even filed an issue about this, and it was promptly closed.)  So yes, perhaps `.ruff.toml` is going to be the most clear compromise.

garrison: I updated this in 746bcddd0beb883090c9c3bd1a491910f2e5225d.

caleb-johnson: Great, thanks

caleb-johnson (approved):

PR #111, opened 2023-04-12
Require 100% test coverage in certain modules

### LINKED ISSUE #107: Adapt coverage testing to require 100% of all new code.
I would not be opposed to explicitly ignoring coverage on wire cutting and EF modules and requiring 100% of everything else implicitly. I feel that much of wire cutting will be replaced by a qpd-based approach soon, and EF was never designed for unit testing anyway. EF is tested with a variety of molecules of different complexities and tested against their known ground state energies. We know the coverage is 85% and barring a large refactor of EF, that is likely where it will always remain.

garrison:
Right now, I have enabled the check for 100% coverage in a few modules that already have it.  Going forward, we plan to also use this check for new modules.  When we introduce new modules that should have 100% coverage, we will have to add their paths to `tox.ini` using globs (e.g., `circuit_knitting_toolbox/circuit_cutting/utils/**/*`), following https://coverage.readthedocs.io/en/7.2.3/source.html#file-patterns.

The old coverage check, of all modules, follows in the coverage workflow, so that we can report the correct number to coveralls.

Closes #107.


garrison:
:point_up: as expected, Coveralls still reports the _overall_ percentage of code coverage.

tox.ini
```diff
@@ -46,6 +46,7 @@ commands =
   coverage3 run --source circuit_knitting_toolbox --parallel-mode -m pytest test/ {posargs}
   coverage3 combine
   coverage3 html
+  coverage3 report --fail-under=100 --show-missing --include="circuit_knitting_toolbox/circuit_cutting/wire_cutting/wire_cutting_verification.py,circuit_knitting_toolbox/utils/integral_driver.py"
```
caleb-johnson, line 49: I was going to approve this, but I think we can remove integral_driver since it's being removed from the repo

garrison: I removed it in 39d197624cc205eb80cd178bd27a9e643b69573d, but we should remember that when we _do_ want to add additional file(s) or path(s), they should be separated by commas.

tox.ini
```diff
@@ -46,6 +46,7 @@ commands =
   coverage3 run --source circuit_knitting_toolbox --parallel-mode -m pytest test/ {posargs}
   coverage3 combine
   coverage3 html
+  coverage3 report --fail-under=100 --show-missing --include="circuit_knitting_toolbox/circuit_cutting/wire_cutting/wire_cutting_verification.py"
```
caleb-johnson, line 49: This looks good. We could even flip it and make 100 the default and point the 85 toward our old tools.

caleb-johnson (approved):
Great, thanks. As I mentioned, I think we could flip the tox coverage setup to make 100 the default and point the 85 toward the older tools.

Up to you, really

garrison:
> I think we _could_ flip the tox coverage setup to make 100 the default and point the 85 toward the older tools.

I like this idea better.  If we don't do this, we're totally going to forget to add a new module to this list at some point.  I've flipped it to an "omission" list in f8e79dda5fc0343c2679b68facdef31d989a6af0.

caleb-johnson (approved):
Nice, looks gtm 👍

PR #113, opened 2023-04-13
Add Docker notebook tests

garrison:


garrison:
Here's an example where the tests failed (before #112 was merged): https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/actions/runs/4684847257/jobs/8301388506

Here's an example where the tests passed (after #112 was merged): https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/actions/runs/4689224708/jobs/8310710653

caleb-johnson (approved):
Would there be any benefit of running this on both ubuntu and windows, due to some of the hardware architecture snafus we've run into?

garrison:
> Would there be any benefit of running this on both ubuntu and windows, due to some of the hardware architecture snafus we've run into?

I suspect not.  The Windows test runners will still have the same hardware as the Ubuntu ones.  It'd be nice if we could test on an M1 Mac, but that is not available, and as long as we have the line added in #71 we should be good.

PR #120, opened 2023-04-28
Temporarily disable qiskit-nature development version test

garrison:
Now that https://github.com/Qiskit/qiskit-nature/pull/1142 has been merged, our "development version tests" are going to [fail](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/actions/runs/4826062552) on `main` until #83 is merged.  This change fixes CI immediately by excluding the development version of qiskit-nature from these tests.  This change should be reverted as part of #83 before it is merged.

PR #121, opened 2023-04-28
Use a released version of the coveralls workflow

garrison:
I noticed when tracking down a [coveralls failure](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/120#issuecomment-1527727777) that we are using the `master` version of the workflow, even though the [README](https://github.com/marketplace/actions/coveralls-github-action) now suggests to use `v2`.  Here, I've updated this.  There's no good reason at this point to rely explicitly on their development version.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -36,7 +36,7 @@ jobs:
       - name: Convert to lcov
         run: coveragepy-lcov --output_file_path coveralls.info
       - name: Upload report to Coveralls
-        uses: coverallsapp/github-action@master
+        uses: coverallsapp/github-action@v2
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


garrison:
I wonder if we'll see a belated coveralls comment like we did at https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/120#issuecomment-1527947818 ...

garrison:
> I wonder if we'll see a belated coveralls comment like we did at [#120 (comment)](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/120#issuecomment-1527947818) ...

I re-ran the workflow on Saturday morning: still no comment.  I ran it again just now, and the comment appeared, at last.

caleb-johnson (approved):
Thanks, hopefully this fixes the intermittent failures

PR #122, opened 2023-05-01
Remove severless deprecations in cutting tutorial

psschwei:
Updates `tutorial_3_cutting_with_quantum_serverless` to incorporate some of the more recent changes made to quantum-serverless:

* use `@distribute_task()` decorator instead of `@run_qiskit_remote()`
* use `with serverless.context():` instead of `with serverless:`

As a result, users will no longer get deprecation warnings  when running the notebook

### DIFF
--- a/docs/tutorials/circuit_cutting/tutorial_3_cutting_with_quantum_serverless.ipynb
+++ b/docs/tutorials/circuit_cutting/tutorial_3_cutting_with_quantum_serverless.ipynb
@@ -127,10 +127,10 @@
     "from typing import Sequence, Any\n",
     "from qiskit import QuantumCircuit\n",
     "from circuit_knitting_toolbox.circuit_cutting.wire_cutting import cut_circuit_wires\n",
-    "from quantum_serverless import run_qiskit_remote, get\n",
+    "from quantum_serverless import distribute_task, get\n",
     "\n",
     "# Create a wrapper function to be sent to remote cluster\n",
-    "@run_qiskit_remote()\n",
+    "@distribute_task()\n",
     "def cut_circuit_wires_remote(\n",
     "    circuit: QuantumCircuit,\n",
     "    method: str,\n",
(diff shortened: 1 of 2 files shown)


psschwei:
@caleb-johnson can you kick the CI for this one?

psschwei:
Looks like there's a version conflict:

```
circuit-knitting-toolbox[notebook-dependencies] 0.0.0 depends on qiskit-terra==0.22.0
quantum-serverless 0.0.7 depends on qiskit-terra>=0.23.3
```

Not sure if we can just bump the qiskit-terra version here, but going to give it a try and see what happens (move fast and break things and all that...)

caleb-johnson:
> Looks like there's a version conflict:
> 
> ```
> circuit-knitting-toolbox[notebook-dependencies] 0.0.0 depends on qiskit-terra==0.22.0
> quantum-serverless 0.0.7 depends on qiskit-terra>=0.23.3
> ```
> 
> Not sure if we can just bump the qiskit-terra version here, but going to give it a try and see what happens (move fast and break things and all that...)

I think it's fine to bump the notebook-dependencies terra version in the context of this PR, assuming that fixes the conflict

garrison:
Looks like now the problem is that quantum-serverless depends on a more recent qiskit-nature than we do.  Might be worth looking at other requirements, too, from https://github.com/Qiskit-Extensions/quantum-serverless/blob/main/client/requirements.txt, e.g., make sure our version of qiskit-ibm-runtime is compatible too with the minimum required by quantum-serverless.

psschwei:
Yeah, comparing against requirements.txt would be the smart way to do it (rather than waiting for the the CI to fail one by one :smile: )

caleb-johnson:
> Yeah, comparing against requirements.txt would be the smart way to do it (rather than waiting for the the CI to fail one by one 😄 )

I think we can get this PR in as-is, but the entanglement forging serverless tutorial probably needs a look as well. If you don't want to do it, we can make an issue.

psschwei:
Having not looked at it and so with no idea what I might be getting myself into, I can tackle the entanglement forging one tomorrow.

caleb-johnson:
> Having not looked at it and so with no idea what I might be getting myself into, I can tackle the entanglement forging one tomorrow.

Haha awesome, there are some problems between serverless and the latest Qiskit Nature classes regarding serialization, but we are still using the old classes in main branch. It should be straightforward, similar to this one

caleb-johnson (approved):
This looks GTM 👍 thanks!

garrison:
Interestly, this PR demonstrates two things about the minimum version tests.

1. They are very nice to have, and nice to run on the notebooks as well.  Without doing so, we probably would have forgotten to bump the version of quantum-serverless.
2. It's a bit unfortunate -- a limitation of the current heuristic for determining the minimum version -- that we have to bump the minimum versions required by the _library_ to match those required transitively by a dependency only of the _notebooks_.

PR #126, opened 2023-05-02
Add release note following Python 3.7 deprecation

garrison:
Python 3.7 was deprecated in #87, but we did not add a release note then because we did not expect to make a release before removing Python 3.7 support.  Now, plans have changed, so we should add the note.


caleb-johnson (approved):
I think this is ready for review? Anyway, looks gtm

PR #135, opened 2023-05-06
Skip slow tests by default (but not in CI)

garrison:
This marks a number of the entanglement forging tests as being "slow."  These slow tests are run by default (and in CI), but they can be excluded by passing the option `--skip-slow` to pytest.  For instance, the command `tox -epy3 -- --skip-slow` runs very quickly.

I am hoping that this helps us iterate more quickly, as the code we are soon to migrate over has a very nice test suite with 100% coverage that executes in only a few seconds.

### DIFF
--- a/test/entanglement_forging/test_entanglement_forging_knitter.py
+++ b/test/entanglement_forging/test_entanglement_forging_knitter.py
@@ -72,7 +72,6 @@ class TestEntanglementForgingKnitter(unittest.TestCase):
 
         return ansatz
 
-    @pytest.mark.slow
     def test_entanglement_forging_H2(self):
         """
         Test to apply Entanglement Forging to compute the energy of a H2 molecule,
@@ -187,7 +186,6 @@ class TestEntanglementForgingKnitter(unittest.TestCase):
         # Ensure ground state energy output is within tolerance
         self.assertAlmostEqual(energy + energy_shift, -75.68366174497027)
 
-    @pytest.mark.slow
     def test_entanglement_forging_driver_H2(self):
         """Test for entanglement forging driver."""
         hcore = np.array([[-1.12421758, -0.9652574], [-0.9652574, -1.12421758]])
--- a/test/entanglement_forging/test_entanglement_forging_knitter.py
+++ b/test/entanglement_forging/test_entanglement_forging_knitter.py
@@ -13,6 +13,8 @@
 
 import os
 import unittest
+import pytest
+
 import numpy as np
 from qiskit.circuit import Parameter, QuantumCircuit
 from qiskit.circuit.library import TwoLocal
@@ -70,6 +72,7 @@ class TestEntanglementForgingKnitter(unittest.TestCase):
 
         return ansatz
 
+    @pytest.mark.slow
     def test_entanglement_forging_H2(self):
         """
         Test to apply Entanglement Forging to compute the energy of a H2 molecule,
@@ -106,6 +109,7 @@ class TestEntanglementForgingKnitter(unittest.TestCase):
         # Ensure ground state energy output is within tolerance
         self.assertAlmostEqual(energy + energy_shift, -1.121936544469326)
 
+    @pytest.mark.slow
     def test_entanglement_forging_H2O(self):  # pylint: disable=too-many-locals
         """
         Test to apply Entanglement Forging to compute the energy of a H20 molecule,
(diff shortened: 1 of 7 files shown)


test/entanglement_forging/test_entanglement_forging_knitter.py
```diff
@@ -70,6 +72,7 @@ def create_mock_ansatz_circuit(self, num_qubits: int) -> QuantumCircuit:
 
         return ansatz
 
+    @pytest.mark.slow
```
caleb-johnson, line 75: I think this one should run in a second or two. I'd be in favor of leaving this test available in every run, just as a sanity check on 80+% of the EF source code.

garrison: Yeah, I should have actually timed these.  I added a timing report in ed173849a3f1e9dbbe19da1ada4d13d508ffaaa2.

Here's what it shows on my current machine:

```
============================= slowest 10 durations =============================
12.03s call     test/entanglement_forging/test_entanglement_forging_knitter.py::TestEntanglementForgingKnitter::test_asymmetric_bitstrings_CN
1.67s call     test/entanglement_forging/test_entanglement_forging_knitter.py::TestEntanglementForgingKnitter::test_asymmetric_bitstrings_O2
1.13s call     test/entanglement_forging/test_entanglement_forging_knitter.py::TestEntanglementForgingKnitter::test_asymmetric_bitstrings_CH3
0.88s call     test/entanglement_forging/test_entanglement_forging_knitter.py::TestEntanglementForgingKnitter::test_entanglement_forging_H2O
0.69s call     test/circuit_cutting/test_circuit_cutting.py::TestCircuitCutting::test_circuit_cutting_manual
0.62s call     test/entanglement_forging/test_entanglement_forging_ground_state_solver.py::TestEntanglementForgingGroundStateSolver::test_entanglement_forging_vqe_hydrogen
0.13s call     test/entanglement_forging/test_entanglement_forging_knitter.py::TestEntanglementForgingKnitter::test_entanglement_forging_H2
0.02s call     test/entanglement_forging/test_entanglement_forging_knitter.py::TestEntanglementForgingKnitter::test_entanglement_forging_driver_H2
```

I enabled the three fastest of these in 27bc0f01bb3e74da6ce53cfa2140f3e8dcba9095.

caleb-johnson: Could we mark O2 and CN as slow and let everything else run? The rest each test slightly different things, so it would be great to have them run in the default as well. CN and O2 are actually redundant and truly don't *need* to be run. If those fail, the CH3 test should fail

garrison: I made these changes, and actually, now coverage reports the same lines covered regardless of whether the slow tests are run.  With this in mind, I decided to flip it such that the slow tests are skipped by default.  All tests are run on CI, or if someone runs `--run-slow` explicitly.  I've also modified it so the report at the end of a test run says _why_ certain tests were skipped:

```
SKIPPED [1] test/circuit_cutting/test_circuit_cutting.py:52: cplex is not installed
SKIPPED [1] test/entanglement_forging/test_entanglement_forging_knitter.py:306: skipping slow test, as --run-slow was not provided
SKIPPED [1] test/entanglement_forging/test_entanglement_forging_knitter.py:231: skipping slow test, as --run-slow was not provided
================== 6 passed, 3 skipped, 453 warnings in 4.48s ==================
```

caleb-johnson: > I made these changes, and actually, now coverage reports the same lines covered regardless of whether the slow tests are run. With this in mind, I decided to flip it such that the slow tests are skipped by default. All tests are run on CI, or if someone runs `--run-slow` explicitly. I've also modified it so the report at the end of a test run says _why_ certain tests were skipped:
> 
> ```
> SKIPPED [1] test/circuit_cutting/test_circuit_cutting.py:52: cplex is not installed
> SKIPPED [1] test/entanglement_forging/test_entanglement_forging_knitter.py:306: skipping slow test, as --run-slow was not provided
> SKIPPED [1] test/entanglement_forging/test_entanglement_forging_knitter.py:231: skipping slow test, as --run-slow was not provided
> ================== 6 passed, 3 skipped, 453 warnings in 4.48s ==================
> ```

Awesome, I realized I was saying "default" before, when I really meant "fast tests only". Glad we can make "fast tests only" the default now :). I suspected CN, CH3, and O2 were testing the exact same pipeline -- glad the coverage backs that up.

caleb-johnson (approved):
Great change, thanks

caleb-johnson:
One random thought, we could run `--run-slow` on a release pre-release workflow if we had tests we'd like to run only as final sanity checks pipeline. Could discuss in a separate issue/PR

PR #138, opened 2023-05-06
Rename wire_cutting package to cutqc

caleb-johnson:
We eventually want to implement wire+gate cutting as extensions of the QPD library. This leaves our current `wire_cutting` package a bit out in the wind.

Packaging `wire_cutting` as cutqc and providing it for users who want the automatic cut finding capability and/or full probability reconstruction still makes sense, but the package should be renamed so as to not create confusion with the primary gate+wire cutting interface.

PR #137, opened 2023-05-06
Initial commit of basic QPD classes, supporting functionality, and unit tests

caleb-johnson:
Resolves #96 
Resolves #97 
Resolves #100 

This PR introduces the basic QPD classes, any required utility functions, and the associated unit tests. I started this PR to give us a place to discuss the packaging of the modules while we finalize a few things internally in the source code.

PR #140, opened 2023-05-06
partition_circuit_qubits

### LINKED ISSUE #101: Create a function for separating a quantum circuit along specified qubit boundaries
To support a circuit cutting workflow, a function is needed which takes a quantum circuit and partitions it according to some boundary inputs.

Resolution of this issue should include a function which automatically replaces instructions which lie across a specified qubit boundary with their optimal decompositions.

caleb-johnson:
Resolves #101 
Resolves #138 

This PR establishes a cutting module in the circuit_cutting package by implementing a `partition_circuit_qubits` function. It assumes cutqc has already been packaged in accordance with #138 , and it assumes the base QPD classes are available as in #137 .

This PR was created so we could discuss the arrangement of the packages, modules, and CI while we finalize the implementation internally.

- [X] Merge #137
- [X] Merge #138 (Need this PR merged because the partition_circuit_qubits tests will take over for the wire_cutting tests in the test_circuit_cutting module)


circuit_knitting_toolbox/circuit_cutting/__init__.py
```diff
@@ -25,4 +25,8 @@
     cutqc.cut_circuit_wires
     cutqc.evaluate_subcircuits
     cutqc.reconstruct_full_distribution
+    partition_circuit_qubits
 """
+from .cutting_decomposition import partition_circuit_qubits
+
+__all__ = ["partition_circuit_qubits"]
```
garrison, line 32: probably worth doing sooner rather than later
```suggestion
__all__ = [
    "partition_circuit_qubits",
]
```

garrison (dismissed):
Thanks.  If the current version passes tests, I think it is good to merge.

garrison (approved):

PR #141, opened 2023-05-07
Create generate_qpd_samples function

### LINKED ISSUE #98: Create a function which allows for easy sampling of multiple QPD bases
In general, a circuit could contain many different elements which could be quasiprobabilistically simulated in one experiment.

Resolution of this issue will include a function which allows for easy sampling of many different QPD bases' joint probability distribution.

caleb-johnson:
Resolves #98

### DIFF
--- a/circuit_knitting_toolbox/circuit_cutting/qpd/qpd.py
+++ b/circuit_knitting_toolbox/circuit_cutting/qpd/qpd.py
@@ -44,7 +44,7 @@ from ...utils.iteration import unique_by_id
 
 
 class WeightType(Enum):
-    """Type of weight."""
+    """Type of weight associated with a QPD sample."""
 
     #: A weight given in proportion to its exact weight
     EXACT = 1
--- a/circuit_knitting_toolbox/circuit_cutting/qpd/qpd.py
+++ b/circuit_knitting_toolbox/circuit_cutting/qpd/qpd.py
@@ -13,7 +13,8 @@
 
 from __future__ import annotations
 
-from collections.abc import Callable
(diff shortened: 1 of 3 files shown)


circuit_knitting_toolbox/circuit_cutting/qpd/qpd.py
```diff
@@ -42,6 +43,63 @@
...
+
+    #: A weight given in proportion to its exact weight
+    EXACT = 1
+
+    #: A weight that was determined through some sampling procedure
+    SAMPLED = 2
+
+
+def generate_qpd_samples(
+    qpd_bases: Sequence[QPDBasis], num_samples: int = 1000
```
garrison, line 57: I wonder if at some point we should remove this default `num_samples`.  It would be so easy to mistakenly write code like
```python
num_samples = 10_000
samples = generate_qpd_samples([gate.basis for gate in qpd_gates])
```

garrison (approved):

circuit_knitting_toolbox/circuit_cutting/qpd/qpd.py
caleb-johnson: Ya it's a good thought. Does the function make sense without this argument being made explicit? I could see it either way.

garrison: I talked to Matthew Treinish about this, and he convinced me that having the default is nice.  It is often convenient to have a reasonable default number of `shots` throughout Qiskit, and `num_samples` is very similar to shots.

PR #146, opened 2023-05-08
Enable a reduced set of pylint checks

garrison:
This is an experiment to see what would have to change in our code base to enable the same set of pylint checks that we've been using internally.  Since pylint can be slow to execute, I've put it as the final check in the `lint` workflow in `tox.ini`.

I wonder if it would be possible to use ruff under nbqa.  (EDIT: yes, see #147.)  If it's possible, would it have caught most of these?

Opening for discussion, mainly.


garrison:
I think we should see what is left after #147 merges, then evaluate whether pylint catches enough that it is worth the extra time during the `lint` workflow.  Adding the "on hold" label until then.

garrison:
Ruff is obviously missing a few things we care about (e.g., #415), so I've revived this.

caleb-johnson (approved):

PR #147, opened 2023-05-08
Enable ruff for jupyter notebooks

garrison:
This was inspired by #146.  There are still a number of errors to fix in the notebooks.  I think we should hold off on this until we migrate all the new notebooks over.

I bumped the nbqa version to 1.6.0 because that is the first version that was advertised to support ruff.


caleb-johnson:
> There are still a number of errors to fix in the notebooks. 
> 

This looks great, thanks. Are the errors you're talking about just the places where you have `# noqa` ?

caleb-johnson (approved):

garrison:
> Are the errors you're talking about just the places where you have `# noqa` ?

That statement is outdated now; yes, all the errors are fixed either by `noqa` or changes in this PR.

PR #153, opened 2023-05-11
Final (0.2.0) cutting workflow

### LINKED ISSUE #104: Create functions which facilitate circuit cutting (decompose, evaluate, reconstruct) at the highest level
Resolution of this issue should include a function which reconstructs the results of the quasiprobability simulation experiments into an expectation value for the full, uncut circuit. This function will need information generated throughout the workflow.

Generally, users should:
 - Decompose their problem into subexperiments #103 
 - Run their subexperiments on a backend
 - Reconstruct the results of the quasiprobability simulation experiments into expectation values -- one for each observable of interest

### LINKED ISSUE #105: Create tutorials showing how to perform gate cutting using CKT

caleb-johnson:
Resolves #104 
Resolves #105 

New tutorials:
- [gate_cutting_to_reduce_circuit_width.ipynb](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/cutting-workflow/docs/circuit_cutting/tutorials/gate_cutting_to_reduce_circuit_width.ipynb)
- [gate_cutting_to_reduce_swap_overhead.ipynb](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/cutting-workflow/docs/circuit_cutting/tutorials/gate_cutting_to_reduce_swap_overhead.ipynb)


docs/circuit_cutting/tutorials/index.rst
```diff
@@ -1,13 +1,9 @@
+#########################
+Circuit Cutting Tutorials
+#########################
 .. include:: README.rst
 
-.. this line is a comment that should be removed once there are tutorials
    nbgallery::
```
garrison, line 6:
```suggestion
.. nbgallery::
```

docs/index.rst
```diff
@@ -15,7 +15,7 @@ The source code to the toolbox is available `on GitHub <https://github.com/Qiski
 
 .. note::
 
-   The `Quantum Serverless <https://qiskit-extensions.github.io/quantum-serverless/>`_ framework is documented separately, as it lives in its own repository.  Check out `Tutorial 2: Forging with Quantum Serverless <./entanglement_forging/tutorials/tutorial_2_forging_with_quantum_serverless.ipynb>`_  and `Tutorial 3: Circuit Cutting with Quantum Serverless <./circuit_cutting/tutorials/tutorial_3_cutting_with_quantum_serverless.ipynb>`_ for examples on how to integrate Quantum Serverless into circuit knitting workflows.
+   The `Quantum Serverless <https://qiskit-extensions.github.io/quantum-serverless/>`_ framework is documented separately, as it lives in its own repository.  Check out `Entanglement Forging Tutorial 2: Forging with Quantum Serverless <./entanglement_forging/tutorials/tutorial_2_forging_with_quantum_serverless.ipynb>`_  and `cutqc utorial 3: Circuit Cutting with Quantum Serverless <./circuit_cutting/tutorials/cutqc/tutorial_3_cutting_with_quantum_serverless.ipynb>`_ for examples on how to integrate Quantum Serverless into circuit knitting workflows.
```
garrison, line 18:
```suggestion
   The `Quantum Serverless <https://qiskit-extensions.github.io/quantum-serverless/>`_ framework is documented separately, as it lives in its own repository.  Check out `Entanglement Forging Tutorial 2: Forging with Quantum Serverless <./entanglement_forging/tutorials/tutorial_2_forging_with_quantum_serverless.ipynb>`_  and `CutQC Tutorial 3: Circuit Cutting with Quantum Serverless <./circuit_cutting/tutorials/cutqc/tutorial_3_cutting_with_quantum_serverless.ipynb>`_ for examples on how to integrate Quantum Serverless into circuit knitting workflows.
```

pyproject.toml
```diff
@@ -76,10 +76,11 @@ docs = [
 notebook-dependencies = [
     "circuit-knitting-toolbox[cplex,pyscf]",
     "quantum-serverless>=0.0.7",
+    "qiskit-nature<0.6.0",
+    "qiskit-ibm-provider",
```
garrison, line 80: I haven't look carefully enough at this yet to understand why it is necessary, but we should put an explicit minimum version here, just like the others.

circuit_knitting_toolbox/entanglement_forging/entanglement_forging_knitter.py
garrison, line 1: Any explanation for this change?  Is it something black did?

pyproject.toml
caleb-johnson: I think it should be removed. I use it in the tutorials, but we should maybe switch to the internal "ExactSampler` in the tutorials when it's ready

circuit_knitting_toolbox/entanglement_forging/entanglement_forging_knitter.py
caleb-johnson: Yes I ran black on `base` conda env by accident and then re-blacked it with the correct version and it spurred this change heh

circuit_knitting_toolbox/circuit_cutting/cutting_evaluation.py
```diff
@@ -82,3 +183,185 @@ def append_measurement_circuit(
...
+            raise ValueError(
+                "Circuit's last register is expected to be named 'observable_measurements'."
+            )
+        if len(circ.cregs) > 1 and circ.cregs[-2].name != "qpd_measurements":
+            raise ValueError(
+                "Circuit's next-to-last register is expected to be named 'qpd_measurements'."
+            )
+
+        if len(circ.cregs) < 2:
+            num_qpd_bits_flat.append(0)  # No gates decomposed to a measurement
```
caleb-johnson, line 285: If we enforce that input circuits cannot contain cregs, I believe we can infer the number of qpd bits during reconstruction by just looking at the size of the corresponding observable. Under this restriction, the number of qpd bits should always be `total_clbits - len(sub-observable)`, right @garrison ?

circuit_knitting_toolbox/circuit_cutting/cutting_evaluation.py
```diff
@@ -82,3 +183,185 @@ def append_measurement_circuit(
...
+    return bases
+
+
+def _get_bases(circuit: QuantumCircuit) -> tuple[list[QPDBasis], list[list[int]]]:
+    """Get a list of each unique QPD basis in the circuit and the QPDGate indices."""
+    bases = []
+    qpd_gate_ids = []
+    for i, inst in enumerate(circuit):
+        if isinstance(inst.operation, SingleQubitQPDGate):
+            raise ValueError(
```
caleb-johnson, line 349: I think this will still be valid for wire cutting. I dont believe there is a use case I'm aware of for wire cutting that doesn't involve separating the circuit

garrison:
Just a quick comment: if 100% coverage for this PR proves to be elusive, I don't think we should get too hung up on it.  I plan to rework the "roundtrip" tests to use this new interface once everything falls into place, and that should bring us up to (or at least near) 100% coverage again.  (It will also cut down the number of lines in those tests, which will be quite nice.)

circuit_knitting_toolbox/circuit_cutting/cutting_evaluation.py
```diff
@@ -14,9 +14,110 @@
...
+    QPDBasis,
+    SingleQubitQPDGate,
+    TwoQubitQPDGate,
+    generate_qpd_samples,
+    decompose_qpd_instructions,
+)
+from .cutting_decomposition import decompose_observables
+
+
+def execute_experiments(
```
caleb-johnson, line 38: @garrison I couldn't get Sphinx to understand the indentions in the args list without adding ticks

circuit_knitting_toolbox/circuit_cutting/cutting_evaluation.py
caleb-johnson: I played with this. I think there is something better we can do, but I'd rather think about it later. It'd be better if the observable measurements and clbits could be added before the midcircuit clbits. That would allow us to use the `cog` in `process_outcomes` to infer which clbits are qpd bits and drop the `num_qpd_bits` arg

docs/circuit_cutting/tutorials/index.rst
```diff
@@ -1,13 +1,9 @@
+#########################
+Circuit Cutting Tutorials
+#########################
 .. include:: README.rst
```
garrison, line 4: The `README.rst` in this directory should be updated, too.  For an example of what it should look like, see `docs/circuit_cutting/cutqc/tutorials/README.rst`.

circuit_knitting_toolbox/circuit_cutting/cutting_decomposition.py
```diff
@@ -86,3 +93,123 @@ def partition_circuit_qubits(
...
+
+    Args:
+        circuit: The circuit containing gates to be decomposed
+        gate_ids: The indices of the gates to decompose
+
+    Returns:
+        A copy of the input circuit with the specified gates replaced with :class:`TwoQubitGate`\ s
+    """
+    # Replace specified gates with TwoQubitQPDGates
+    new_qc = circuit.copy()
```
garrison, line 110: Thoughts on adding an `inplace` kwarg?  It can default to `False`, and then we would get the current behavior.  Not urgent; we can always do this later if we wish.

circuit_knitting_toolbox/circuit_cutting/cutting_decomposition.py
```diff
@@ -86,3 +93,123 @@ def partition_circuit_qubits(
...
+        decomposition = QPDBasis.from_gate(gate.operation)
+        qpd_gate = TwoQubitQPDGate(decomposition, label=f"cut_{gate.operation.name}")
+        new_qc.data[gate_id] = CircuitInstruction(qpd_gate, qubits=qubit_indices)
+
+    return new_qc
+
+
+def partition_problem(
+    circuit: QuantumCircuit,
+    partition_labels: Sequence[str | int],
```
garrison, line 123: I believe we decided at some point that `str | int` should be replaced with `Hashable`.  (I will make this comment only here, but there are 11 instances throughout this PR.)

caleb-johnson: Not all hashables are sortable, and in order to be consistent with ordering from our subcircuit/subobservable dicts,  we sort by partition name often.

I think it's fine for the underlying package to be more general and for us to restrict the inputs through our interface for this reason, but open to suggestions

caleb-johnson: Sorting will also break if they pass in a mix of strings and ints

circuit_knitting_toolbox/circuit_cutting/cutting_evaluation.py
```diff
@@ -14,9 +14,112 @@
...
+    circuits: QuantumCircuit | dict[str | int, QuantumCircuit],
+    observables: PauliList | dict[str | int, PauliList],
+    num_samples: int,
+    samplers: BaseSampler | Sequence[BaseSampler] | None = None,
+) -> tuple[list[list[list[tuple[QuasiDistribution, int]]]], Sequence[float]]:
```
garrison, line 46: It'd be nice to use `namedtuple` for these functions that return multiple things (i.e., this and `partition_problem`).  Not urgent, as this change will be backwards compatible.

docs/circuit_cutting/tutorials/gate_cutting_to_reduce_circuit_width.ipynb
```diff
@@ -0,0 +1,239 @@
...
+    "    [\"ZZIIIII\", \"IZZIIII\", \"IIZZIII\", \"IIIIXXI\", \"IIIIXXI\", \"IIIIIXX\"]\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "162a5629",
+   "metadata": {},
+   "source": [
+    "#### Decompose the circuit and observables according to a specified qubit partitioning using `decompose_problem`\n",
```
garrison, line 89:
```suggestion
    "#### Decompose the circuit and observables according to a specified qubit partitioning using `partition_problem`\n",
```

docs/circuit_cutting/tutorials/gate_cutting_to_reduce_circuit_width.ipynb
```diff
@@ -0,0 +1,239 @@
...
+    "\n",
+    "*Arguments*\n",
```
garrison, line 91: IMO, the explicit listing of all inputs and outputs distracts from the prose/flow.  But I kind of like where you are going with this, having them lead the discussion.  Here's what I'd do differently: for the inputs, just describe them in a sentence.  Then, elaborate on the input variables that have not been explained yet.  For instance:

> Decompose the circuit and observables according to a specified qubit partitioning using `partition_problem`.  This function takes the input `circuit`, a sequence of `observables`, and a sequence we call `partition_labels`, which describes which partition each input qubit belongs to.  `partition_labels` is a sequence of labels, one for each qubit in the circuit.  Qubits with the same label will be grouped together in the same partition.  Operations that span qubits in multiple partitions will be cut.

I think that a bulleted list is fine for the outputs (when the output is a tuple), but I'd put words about outputs _after_ the relevant code block.

garrison: > Then, elaborate on the input variables that have not been explained yet.

If there are many, a bulleted list might still make sense.  But I think it makes sense to skip over explaining input variables whose explanation is obvious or redundant with what came earlier.

docs/circuit_cutting/tutorials/gate_cutting_to_reduce_circuit_width.ipynb
```diff
@@ -0,0 +1,239 @@
...
+   "id": "1e710888",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "from qiskit.circuit.library import EfficientSU2\n",
+    "from qiskit.quantum_info import PauliList\n",
+    "from qiskit.primitives import Estimator\n",
+    "\n",
+    "from circuit_knitting_toolbox.utils.simulation import ExactSampler as Sampler\n",
```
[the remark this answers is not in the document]
garrison: If we do this, we might as well use Aer's Estimator, too, for comparison.

docs/circuit_cutting/tutorials/gate_cutting_to_reduce_swap_overhead.ipynb
```diff
@@ -0,0 +1,301 @@
...
+   "source": [
+    "transpiled_qc.draw(\"mpl\", scale=0.4, idle_wires=False)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "6c106db7",
+   "metadata": {},
+   "source": [
+    "#### Replace gates resulting in swaps with TwoQubitQPDGates by specifying their indices using `decompose_gates`\n",
```
garrison, line 144:
```suggestion
    "#### Replace gates resulting in swaps with `TwoQubitQPDGate`s by specifying their indices using `decompose_gates`\n",
```

circuit_knitting_toolbox/circuit_cutting/cutting_reconstruction.py
```diff
@@ -13,10 +13,78 @@
...
+        subobservables = observables
+        subsystem_observables = {
+            label: ObservableCollection(so) for label, so in observables.items()
+        }
+        expvals = np.zeros(len(list(observables.values())[0]))
+
+    # Assign each weight's sign and calculate the expectation values for each observable
+    for i, _ in enumerate(quasi_dists):
+        sorted_subsystems = sorted(subsystem_observables.keys())  # type: ignore
+        coeff = coefficients[i]
```
garrison, line 66:
```suggestion
    for i, coeff in enumerate(coefficients):
        sorted_subsystems = sorted(subsystem_observables.keys())  # type: ignore
```

circuit_knitting_toolbox/circuit_cutting/cutting_decomposition.py
caleb-johnson: Done. Added one for partition_circuit_qubits as well

circuit_knitting_toolbox/circuit_cutting/cutting_decomposition.py
```diff
@@ -83,6 +92,132 @@ def partition_circuit_qubits(
...
+    qpd_circuit = partition_circuit_qubits(circuit, partition_labels)
+    i = 0
+    for inst in qpd_circuit.data:
+        if isinstance(inst.operation, TwoQubitQPDGate):
+            inst.operation.label = inst.operation.label + f"_{i}"
+            i += 1
+
+    # Separate the decomposed circuit into its subcircuits
+    qpd_circuit_dx = qpd_circuit.decompose(TwoQubitQPDGate)
+    separated_circs = separate_circuit(qpd_circuit_dx)
```
garrison, line 181:
```suggestion
    separated_circs = separate_circuit(qpd_circuit_dx, partition_labels)
```
otherwise the labels get their default: 0, 1, ...
and with this change, `subcircuits` and `subobservables` will have the same keys in the tutorial.

circuit_knitting_toolbox/circuit_cutting/cutting_decomposition.py
garrison: I am still hoping that it's possible to remove the sort and broaden this type.  I will follow up more here when I better understand the implications of this.

docs/circuit_cutting/tutorials/gate_cutting_to_reduce_circuit_width.ipynb
```diff
@@ -0,0 +1,239 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "id": "30326299",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "subcircuits, subobservables = partition_problem(\n",
+    "    circuit, \"AAABBBB\", observables=observables\n",
+    ")"
```
garrison, line 112: I think it'd be very instruction after this to have three cells containing these outputs:

- `subobservables`
- `subcircuits["A"].draw("mpl")` (once https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/153/files#r1198084353 is changed)
- `subcircuits["B"].draw("mpl")`

circuit_knitting_toolbox/circuit_cutting/cutting_evaluation.py
```diff
@@ -82,3 +185,184 @@ def append_measurement_circuit(
...
+    coefficients = []
+    sampled_frequencies = []
+    for z, (map_ids, sample_weight) in enumerate(sorted_samples):
+        redundancy = sample_weight[0]
+        subexperiments.append([])
+        actual_coeff = np.prod(
+            [basis.coeffs[map_id] for basis, map_id in strict_zip(bases, map_ids)]
+        )
+        sampled_coeff = (redundancy / num_samples) * (kappa * np.sign(actual_coeff))
+        coefficients.append(sampled_coeff)
```
garrison, line 244: Some day, we're also going to want the weight type of each coefficient later in the workflow.  I'm thinking something like

```suggestion
    for z, (map_ids, (redundancy, weight_type)) in enumerate(sorted_samples):
        subexperiments.append([])
        actual_coeff = np.prod(
            [basis.coeffs[map_id] for basis, map_id in strict_zip(bases, map_ids)]
        )
        sampled_coeff = (redundancy / num_samples) * (kappa * np.sign(actual_coeff))
        coefficients.append((sampled_coeff, weight_type))
```

Of course, this will need to be coupled with a change when the `coefficients` are processed later in the workflow.

If we do this now, we won't have to break this interface later.  But it's not strictly necessary to do this until we want to be able to propagate error bars.

(It's too bad the diff is not more compact; I edited the first and last lines, while removing the second.)

docs/circuit_cutting/tutorials/gate_cutting_to_reduce_swap_overhead.ipynb
```diff
@@ -0,0 +1,301 @@
...
+       "<Figure size 1634.66x294.311 with 1 Axes>"
+      ]
+     },
+     "execution_count": 6,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "qpd_circuit = decompose_gates(circuit, [8, 20, 32])\n",
```
garrison, line 173: It might be more instructive to programmatically generate these indices than to pull them out of thin air.

```suggestion
    "cut_indices = [\n",
    "    i for i, instruction in enumerate(circuit.data)\n",
    "    if {circuit.find_bit(q)[0] for q in instruction.qubits} == {0, 3}\n",
    "]\n",
    "qpd_circuit = decompose_gates(circuit, cut_indices)\n",
```

docs/circuit_cutting/tutorials/gate_cutting_to_reduce_swap_overhead.ipynb
```diff
@@ -0,0 +1,301 @@
...
+    "qpd_circuit.draw(\"mpl\", scale=0.8)"
+   ]
+  },
+  {
```
garrison, line 177: It'd be quite nice, right here, to pass `qpd_circuit` through the transpiler, at least to the stage that a layout is determined.  For this to work as expected, though, the circuit will need to be decomposed all the way to `SingleQubitQPDGate`s.  Otherwise, the routing layer will want to add a bunch of swaps to get the two cut qubits on adjacent qubits :joy_cat:.

I don't know that this is a pressing priority, but it would be very nice for this demonstration.  The motivation at the top of this tutorial shows an example where transpilation results in swap overhead.  For symmetry, it'd be nice to see the layout and routing with _reduced_ swap overhead.

circuit_knitting_toolbox/circuit_cutting/cutting_evaluation.py
```diff
@@ -14,9 +14,110 @@
...
+    decompose_qpd_instructions,
+)
+from .cutting_decomposition import decompose_observables
+
+
+def execute_experiments(
+    circuits: QuantumCircuit | dict[str | int, QuantumCircuit],
+    observables: PauliList | dict[str | int, PauliList],
+    num_samples: int,
+    samplers: BaseSampler | Sequence[BaseSampler] | None = None,
```
garrison, line 43: As I mentioned elsewhere, I think it would be better to get rid of this default.  If the user actually wants `ExactSampler`, they should provide it explicitly.  It's not something that should happen by accident.

circuit_knitting_toolbox/circuit_cutting/cutting_decomposition.py
```diff
@@ -83,6 +92,132 @@ def partition_circuit_qubits(
...
+
+    subobservables_by_subsystem = {
+        label: observables_restricted_to_subsystem(qubits, observables)
+        for label, qubits in qubits_by_subsystem.items()
+    }
+
+    subsystem_observables = {
+        label: ObservableCollection(subobservables)
+        for label, subobservables in subobservables_by_subsystem.items()
+    }
```
garrison, line 221: I think we should remove this and return only `subobservables_by_subsystem`.  It's needed in two places in the code, and in those places we can just compute it like this.  In fact, in `cutting_reconstruction.py`, a comprehension like this already appears in one of the branches.

docs/circuit_cutting/tutorials/gate_cutting_to_reduce_circuit_width.ipynb
```diff
@@ -0,0 +1,239 @@
...
+    "%%capture\n",
+    "\n",
+    "sampler1 = Sampler()\n",
+    "sampler2 = Sampler()\n",
+    "\n",
+    "quasi_dists, coefficients = execute_experiments(\n",
+    "    subcircuits,\n",
+    "    subobservables,\n",
+    "    num_samples=1500,\n",
+    "    samplers=[sampler1, sampler2],\n",
```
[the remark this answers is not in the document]
caleb-johnson: I still think it should be called `samplers`, but otherwise I implemented this as you describe here

circuit_knitting_toolbox/circuit_cutting/cutting_decomposition.py
```diff
@@ -38,36 +42,38 @@ def partition_circuit_qubits(
...
+        qubit_indices = [circuit.find_bit(qubit).index for qubit in instruction.qubits]
         partitions_spanned = {partition_labels[idx] for idx in qubit_indices}
         # Ignore local gates and gates that span only one partition
         if (
             len(qubit_indices) <= 1
             or len(partitions_spanned) == 1
             or isinstance(instruction.operation, Barrier)
         ):
-            continue
+            continue  # pragma: no cover
```
garrison, line 76: I commented on this line in the commit, but I don't see it now.  In any case ... I just realized that this function is not meaningfully changed in this PR.  In fact, I've been seeing coverage claim that this line is missing coverage locally, even though it reaches 100% coverage in CI.  I suppose you must be experiencing this, too?  It's odd, because I swear that this line _must_ be hit -- we test things like barriers, etc., right?

circuit_knitting_toolbox/circuit_cutting/cutting_reconstruction.py
```diff
@@ -13,10 +13,82 @@
...
+from qiskit.result import QuasiDistribution
 
-from ..utils.observable_grouping import CommutingObservableGroup
+from ..utils.observable_grouping import CommutingObservableGroup, ObservableCollection
 from ..utils.bitwise import bit_count
+from .cutting_decomposition import decompose_observables
+from .qpd import WeightType
+
+
+def reconstruct_expectation_values(
```
caleb-johnson, line 28: Need input checking on this function. Make sure inputs are compatible

circuit_knitting_toolbox/circuit_cutting/cutting_evaluation.py
```diff
@@ -14,9 +14,113 @@
...
+    WeightType,
+)
+from .cutting_decomposition import decompose_observables
+
+
+def execute_experiments(
+    circuits: QuantumCircuit | dict[str | int, QuantumCircuit],
+    observables: PauliList | dict[str | int, PauliList],
+    num_samples: int,
+    samplers: BaseSampler | dict[str | int, BaseSampler],
```
caleb-johnson, line 43: We can actually handle `AerSampler` with `shots=None`. We just check for that specific case and switch to `ExactSampler`

docs/circuit_cutting/tutorials/gate_cutting_to_reduce_circuit_width.ipynb
garrison, line 1: It'd be nice if the tutorials calculate and show the product of the `overhead` from each `QPDBasis`.  This is a pretty important quantity, and we should draw attention to it.  Otherwise, people may not understand why their results are "wrong."

circuit_knitting_toolbox/circuit_cutting/cutting_evaluation.py
garrison: My first reaction was :laughing:.  But it's a reasonable idea.  If we're going to do this, we should give the user a warning as well.

caleb-johnson: I implemented this, and I like the behavior.

Users passing in qiskit.primitives.Sampler or qiskit_aer.primitives.Sampler(shots=None) are explicitly asking for exact sub-expectation values, so it's not too far of a leap to default to a class which can do that for them.

README.md
```diff
@@ -4,7 +4,7 @@
   [![Stability](https://img.shields.io/badge/Stability-alpha-f4d03f.svg)](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/releases)
   ![Platform](https://img.shields.io/badge/Platform-Linux%20%7C%20macOS-informational)
   [![Python](https://img.shields.io/badge/Python-3.8%20%7C%203.9%20%7C%203.10%20%7C%203.11-informational)](https://www.python.org/)
-  [![Qiskit](https://img.shields.io/badge/Qiskit-%E2%89%A5%200.42.1-6133BD)](https://github.com/Qiskit/qiskit)
+  [![Qiskit](https://img.shields.io/badge/Qiskit-%E2%89%A5%200.43.0-6133BD)](https://github.com/Qiskit/qiskit)
```
caleb-johnson, line 7: Allows us to more easily inspect and handle the `options` field of a `Sampler`. This will continue making our life easier as we develop around the Sampler, we should just bump it now

docs/circuit_cutting/tutorials/gate_cutting_to_reduce_circuit_width.ipynb
caleb-johnson: Done in both tutorials. This involved returning `bases` from both `decompose_problem` in tutorial 1 and `decompose_gates` in tutorial 2. I think this is totally reasonable. Like you stated, this is an important value, and it should be made easily available.

If I didn't return `bases` from `decompose_problem`, it would be a fair hassle for the user to find all the unique bases across all their subcircuits.

docs/circuit_cutting/tutorials/gate_cutting_to_reduce_swap_overhead.ipynb
caleb-johnson: In tutorial 2, I have included a cell showing a specific decomposition of the QPD circuit and how its depth is less than half the depth of the original circuit after they are both transpiled.

caleb-johnson:
> Just a quick comment: if 100% coverage for this PR proves to be elusive, I don't think we should get too hung up on it. I plan to rework the "roundtrip" tests to use this new interface once everything falls into place, and that should bring us up to (or at least near) 100% coverage again. (It will also cut down the number of lines in those tests, which will be quite nice.)

I've ignored the cutting modules in coverage for now. I will leave this comment open and try to get 100% coverage by release date. Worst case, we get as close as we can and leave the coverage ignore on until next release.

docs/circuit_cutting/tutorials/gate_cutting_to_reduce_circuit_width.ipynb
caleb-johnson: I've cleaned up the explanation, but I didn't put the output explanation after the cell blocks. I'm still open to it, but I think I prefer it this way

circuit_knitting_toolbox/circuit_cutting/cutting_decomposition.py
caleb-johnson: I'm going to resolve this because I think it is out of scope for this release.

That being said, what we have now is *more* strict than you think is necessary, so loosening this to all `Hashable` in a future release shouldn't break anything.

docs/circuit_cutting/tutorials/gate_cutting_to_reduce_circuit_width.ipynb
caleb-johnson: Going to resolve this, we can continue to refine the flow leading up to release

docs/circuit_cutting/tutorials/gate_cutting_to_reduce_circuit_width.ipynb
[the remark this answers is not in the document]
caleb-johnson: Fun fact, the Aer estimator was a source of error in our testing last week. When switching over to that estimator, I needed to set a couple of options to get it to return exact expvals; whereas, terra estimator was returning the exact by default.

Now that you've implemented exact weighting, we have access to exact sub-expvals, and the estimator isn't contributing to the error, users can simulate exact expectation values in the tutorials and then begin running things in noisier ways (real backends, bigger number of cuts, etc).

To me, being able to simulate the expvals exactly is an intuitive starting point for *learning* the technique. Being able to move to a noisy backend and observe the change in quality of expval is helpful. Also, being able to see how the exact/sampled weights affect the answer as `num_samples` moves is also instructive.

Really pleased with the `Sampler` interface overall right now, including shadow-support for AerSampler/no shots and TerraSampler

caleb-johnson:
Need to link to API docs for all CKT class references in cutting tutorials

README.md
garrison: The only downside I see with bumping it now is that, then, there is no Qiskit version supported that will run without deprecation warnings (#161).  But, since this is alpha software, I think this is okay.

caleb-johnson: I am going to do the best I can with those warnings for 0.3.0. I believe updating EntForging to latest Qiskit and Qiskit Nature will resolve 95%+ of these

caleb-johnson:
> Just a quick comment: if 100% coverage for this PR proves to be elusive, I don't think we should get too hung up on it. I plan to rework the "roundtrip" tests to use this new interface once everything falls into place, and that should bring us up to (or at least near) 100% coverage again. (It will also cut down the number of lines in those tests, which will be quite nice.)

We have 100% coverage after reworking the roundtrip

caleb-johnson:
> Need to link to API docs for all CKT class references in cutting tutorials (i.e. `QPDBasis`, `TwoQubitQPDGate`)

@garrison should this be done in a follow PR? I don't think the Sphinx docs will be live until this is merged to main, so the links would be dead until that time.

garrison:
> > Need to link to API docs for all CKT class references in cutting tutorials (i.e. `QPDBasis`, `TwoQubitQPDGate`)
> 
> @garrison should this be done in a follow PR? I don't think the Sphinx docs will be live until this is merged to main, so the links would be dead until that time.

Makes sense to me.

docs/circuit_cutting/explanation/index.rst
```diff
@@ -1,5 +1,43 @@
-###########################
-Circuit Cutting Explanation
-###########################
+###################################################
```
caleb-johnson, line 1: @garrison I think the QPD explanatory can/should go in here, rather than having an extra explanatory. We can just have a QPD section after the overview I think

circuit_knitting_toolbox/circuit_cutting/qpd/qpd.py
```diff
@@ -384,23 +384,24 @@ def _decompose_qpd_measurements(
...
-    circuit.add_register(reg)
-
-    # Place the measurement instructions
-    for idx, i in enumerate(qpd_measure_ids):
-        gate = circuit.data[i]
-        inst = CircuitInstruction(
-            operation=Measure(), qubits=[gate.qubits], clbits=[reg[idx]]
-        )
-        circuit.data[i] = inst
+    if len(qpd_measure_ids) > 0:
```
caleb-johnson, line 387: I snuck this change in, but I want to draw attention to it.

Let's not add empty QPD registers if the circuit doesn't have any mid-circuit measurements. This came up when I was writing tests for the cutting code. I think it's more intuitive for a circuit to not have a qpd register if no qpd measurements are expected

test/circuit_cutting/qpd/test_qpd.py
```diff
@@ -104,15 +104,13 @@ def test_decompose_qpd_instructions(self):
         with self.subTest("Empty circuit"):
             circ = QuantumCircuit()
             new_circ = decompose_qpd_instructions(QuantumCircuit(), [])
-            circ.add_register(ClassicalRegister(0, name="qpd_measurements"))
```
caleb-johnson, line 107: These changes go with the earlier comment regarding adding empty QPD registers to subexperiments when no QPD measurements are present

test/circuit_cutting/test_cutting_reconstruction.py
```diff
@@ -12,13 +12,17 @@
...
 import numpy as np
+from qiskit.result import QuasiDistribution
 from qiskit.quantum_info import Pauli, PauliList
 from qiskit.circuit import QuantumCircuit, ClassicalRegister
 
 from circuit_knitting_toolbox.utils.observable_grouping import CommutingObservableGroup
+from circuit_knitting_toolbox.circuit_cutting.qpd import WeightType
 from circuit_knitting_toolbox.circuit_cutting.cutting_reconstruction import (
-    process_outcome,
+    _process_outcome,
```
caleb-johnson, line 24: I made process_outcome and append_circuit_measurements private for 0.2.0, as that is the style I used in the rest of the modules. I wanted to keep all helper functions private until it was clear they would be a value-add to the API.

I want to keep the initial API as light as possible

circuit_knitting_toolbox/circuit_cutting/cutting_evaluation.py
```diff
@@ -77,8 +200,252 @@ def append_measurement_circuit(
...
+                "SingleQubitQPDGates are not supported in unseparable circuits."
+            )
+        if isinstance(inst.operation, TwoQubitQPDGate):
+            bases.append(inst.operation.basis)
+            qpd_gate_ids.append([i])
+
+    return bases, qpd_gate_ids
+
+
+def _validate_samplers(
```
garrison, line 396: As I think about this more, I am becoming concerned about this business of intercepting the sampler and replacing it with a sampler that is different than the one the user provided.  If this going to be done, it should be done in a way that the user is informed what is happening, but this is difficult for two reasons:

1. A warning will only be displayed once in a given Python session
2. In the notebooks, we often use `%%capture` in cells that call the samplers, so the warning never makes it to the user.

I think it is far better to try to use the provided sampler and to fail with an informative error message, if possible, if the sampler is not suitable for our us.  We could create a small "canary" circuit, which we run if provided a Terra or Aer sampler.  If we run this circuit and it either fails or gives an unexpected result, we inform the user what failed and provide a suggestion.  Otherwise, we can go forward assuming the user is now on a more recent version of Terra or Aer, where the issues we've found with these samplers have been fixed.

If we want, for user convenience, we could even have a `def sampler_factory(*, shots=1024)` function in `simulation.py`, which would return an appropriately configuredAer sampler if `shots is not None` and `ExactSampler` otherwise.  In the future, when Aer's sampler supports everything we need it to, we can deprecate this and suggest people use Aer's sampler directly instead.  This convenience function would exist to allow the user to adjust one parameter (`shots`) and get the most appropriate sampler in every situation.

Overall, I think it is much better for us to have a dialogue with the user than do something that is beyond their control and likely beyond their awareness.  If a user is investigating something unexpected, they might try with different samplers, but if we're swapping out the sampler behind the scenes, this is wasted effort.  Further, being honest about the limitations of current samplers is probably good for us: it raises awareness of these issues rather than sweeping them under the rug.

test/circuit_cutting/test_cutting_reconstruction.py
garrison: Yes, I saw this, and I agree this makes most sense.

circuit_knitting_toolbox/circuit_cutting/qpd/qpd.py
garrison: OK, yeah, I noticed this and I am glad you brought this up.  I am not sure I agree -- I think that having the register be there consistently is more in line with expectation.  I also don't think that this allowed the tests to be simplified all that much (I think you mentioned earlier that this was part of the motivation for doing this?)

circuit_knitting_toolbox/circuit_cutting/cutting_evaluation.py
caleb-johnson: OK, let's just raise a `ValueError` and tell users to use `circuit_knitting_toolbox.utils.ExactSampler` if they want exact results.

Then the question is, what do we use in the tutorials? I like the exact expvals in tutorials. If we're going to have undeterministic results in tutorials, I'd rather run a bunch of experiments and show how the results are scattered around the true mean or something.

garrison: > let's just raise a `ValueError`

I think something like `RuntimeError` is more appropriate, but I am interested to hear your reasoning.

> Then the question is, what do we use in the tutorials?

Yeah, that was in part the motivation behind my `sampler_factory` function.  Then we could use `sampler_factory(shots=None)` in the tutorials, and if the user actually wanted to have a finite number of shots, they could adjust it right there and then the aer sampler would be used instead.

caleb-johnson: I think I'd rather just error out if the user passes an exact sampler that doesnt support midcirc meas. We can inform them exactly why their sampler doesnt work and how they can fix it (user ExactSampler).

Like you said, ExactSampler is non-physical, and I see it more of a debugging/learning tool. I think we should continue allowing users to directly set up their Sampler and pass it in. If we do the factory, it's going to need runtime info and options be passed in, and I'm sure there is other complexity. I'd rather not introduce that as a step.

garrison: > If we do the factory, it's going to need runtime info and options be passed in, and I'm sure there is other complexity. I'd rather not introduce that as a step.

If we do the factory, I was thinking it should expose a very simple API, where `shots` is the only thing you can vary.  If you need anything more complex than that, use the appropriate class directly.

That said, I'm not certain this would be more clear if the class is hidden out in `simulation.py`.  In fact, maybe the best way to do this is to define it **directly in the notebook**, rather than in `simulation.py`:

```python
def best_sampler(shots: int = 1024) -> BaseSampler:
    if shots is None:
        return ExactSampler()
    else:
        return AerSampler(shots=shots)  # or whatever is necessary to set the shots count correctly
```

caleb-johnson: OK let's do this in the notebook and write a little comment specifying why it is necessary

circuit_knitting_toolbox/circuit_cutting/cutting_evaluation.py
```diff
@@ -77,8 +203,240 @@ def append_measurement_circuit(
         actual_qubit = qubit_locations[subqubit]
         if genobs_x[subqubit]:
             if genobs_z[subqubit]:
-                qc.sdg(actual_qubit)
+                qc.sdg(actual_qubit)  # pragma: no coverage
```
caleb-johnson, line 206: @garrison not urgent, but do you have an immediate idea of how to trigger this? Can we add something to `test_cutting_roundtrip`?

garrison: You need an observable with a Pauli `Y` in it somewhere.

garrison: Oh wait.  There _is_ such an observable.  Hmm.

caleb-johnson: The roundtrip observable is "ZZZ". Let me add a Y

garrison: There were such observables in the roundtrip tests, but it looks like you eliminated them.  I chose those observables to have a good mix of I,X,Y,Z and different phase coefficients.

caleb-johnson: Fixed

caleb-johnson:
Need a release note for this PR

circuit_knitting_toolbox/circuit_cutting/cutting_evaluation.py
```diff
@@ -82,3 +208,234 @@ def append_measurement_circuit(
...
+        if isinstance(inst.operation, TwoQubitQPDGate):
+            bases.append(inst.operation.basis)
+            qpd_gate_ids.append([i])
+
+    return bases, qpd_gate_ids
+
+
+def _validate_samplers(samplers: BaseSampler | dict[str | int, BaseSampler]) -> None:
+    """Replace unsupported statevector-based Samplers with ExactSampler."""
+    samplers_out = copy.copy(samplers)
```
garrison, line 391: should be able to get rid of `samplers_out` at this point

circuit_knitting_toolbox/circuit_cutting/cutting_reconstruction.py
```diff
@@ -13,13 +13,93 @@
...
+    quasi_dists: Sequence[Sequence[Sequence[tuple[QuasiDistribution, int]]]],
+    coefficients: Sequence[tuple[float, WeightType]],
+    observables: PauliList | dict[str | int, PauliList],
```
garrison, line 31: fyi, we'll need both `observables` and `subobservables` here to support `SparsePauliOp`.

circuit_knitting_toolbox/circuit_cutting/cutting_evaluation.py
```diff
@@ -13,13 +13,139 @@
...
+
+    Returns:
+        A tuple containing a 3D list of length-2 tuples holding the quasi-distributions
+        and QPD bit information for each sub-experiment as well as the coefficients corresponding
+        to each unique subexperiment's sampling frequency.
+
+    Raises:
+        ValueError: The number of requested samples must be positive.
+        ValueError: The types of ``circuits`` and ``observables`` arguments are incompatible.
+        ValueError: ``SingleQubitQPDGate``\ s are not supported in unseparable circuits.
```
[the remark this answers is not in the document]
caleb-johnson: Off the top of my head, the logic gets more complex with single qubit gates, because you need to look whether they're part of a greater decomposition and do everything in sync. Let me look closer for specifics

circuit_knitting_toolbox/circuit_cutting/cutting_evaluation.py
```diff
@@ -82,3 +208,234 @@ def append_measurement_circuit(
...
+    """Run subexperiments on the backend."""
+    num_qpd_bits_flat = []
+
+    # Run all the experiments in one big batch
+    experiments_flat = list(chain.from_iterable(subexperiments))
+    for circ in experiments_flat:
+        assert not (
+            len(circ.cregs) < 1 or circ.cregs[-1].name != "observable_measurements"
+        )
+        assert not (len(circ.cregs) > 1 and circ.cregs[-2].name != "qpd_measurements")
```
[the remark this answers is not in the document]
caleb-johnson: OK, I have been meaning to consider this. This was my top option too.

Check whether the input to `decompose` step has 0 cregs, and check whether input to `execute` step has 1 creg named `qpd_measurements`.

releasenotes/notes/gate-cutting-workflow-ebbedbdb1313b258.yaml
```diff
@@ -0,0 +1,34 @@
+---
+prelude: >
+    0.2.0 is centered around the addition of functions which allow for the easy
+    implementation of gate cutting workflows.
+
+    The foundation of the CKT circuit cutting API is the :mod:`circuit_knitting_toolbox.circuit_cutting.qpd`
+    module. This module allows for easy transformation of elements of quantum circuits (gates and wires) into
```
caleb-johnson, line 7: "module" --> "package"

circuit_knitting_toolbox/circuit_cutting/cutting_reconstruction.py
garrison: I said earlier that we will need both `subobservables` and `observables` for reconstruction with `SparsePauliOp`, but this is not sufficient, I am now thinking.  We actually need `partition_labels` again, too, so that we know how each observable matches up with the different subobservables.  But at that point, we can re-derive the `subobservables` from `observables` and `partition_labels`.  (Of course, if the circuit is not separated, we don't actually need `partition_labels` because it is trivial.)

circuit_knitting_toolbox/circuit_cutting/cutting_evaluation.py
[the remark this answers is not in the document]
caleb-johnson: OK I do input checks on both `decompose_gates` and `partition_problem`. I also do checks on all the circuits before executing them in batches in `_run_experiments_batch`

circuit_knitting_toolbox/circuit_cutting/cutting_reconstruction.py
caleb-johnson: OK what do you suggest we do for 0.2.0? Just hold it as it is for now and rename this to `subobservables`?

garrison: For `execute`, I think it makes sense to use `subobservables`.  For reconstruction, I think `observables` actually makes more sense.

circuit_knitting_toolbox/circuit_cutting/cutting_evaluation.py
```diff
@@ -13,13 +13,139 @@
...
+)
+from .cutting_decomposition import decompose_observables
 
-def append_measurement_circuit(
+
+class ExperimentResults(NamedTuple):
+    """The results of running each sampled subexperiment on the backend."""
+
+    quasi_dists: list[list[list[tuple[QuasiDistribution, int]]]]
+    coeffs: Sequence[tuple[float, WeightType]]
```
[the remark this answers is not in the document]
garrison: lgtm

circuit_knitting_toolbox/circuit_cutting/cutting_evaluation.py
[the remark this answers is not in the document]
caleb-johnson: We will leave this error for 0.2.0. Let's see how everything falls out with wire cutting and 0.3.0 and decide whether this is still a question

docs/circuit_cutting/explanation/index.rst
```diff
@@ -1,5 +1,46 @@
...
+Explanatory material for the circuit cutting module
+###################################################
 
-Circuit cutting is a technique to decompose a quantum circuit into smaller circuits, whose results can be knitted together to reconstruct the original circuit output.
+Overview of circuit cutting
+---------------------------
+Circuit cutting is a technique to increase the size of circuits we can run on quantum hardware at the cost of an additional sampling overhead. A larger quantum circuit can be decomposed by cutting its gates and wires into smaller circuits, which can be executed within the constraints of available quantum hardware. The results of these smaller circuits are combined to reconstruct the outcome of the original problem. Circuit cutting can also be used to engineer gates between distant qubits which would otherwise require a large swap overhead.
```
garrison, line 7: Just a suggestion
```suggestion
Circuit cutting is a technique to increase the size of circuits we can run on quantum hardware at the cost of an additional sampling overhead. A larger quantum circuit can be decomposed by cutting its gates and/or wires, resulting in smaller circuits which can be executed within the constraints of available quantum hardware. The results of these smaller circuits are combined to reconstruct the outcome of the original problem. Circuit cutting can also be used to engineer gates between distant qubits which would otherwise require a large swap overhead.
```

docs/circuit_cutting/explanation/index.rst
```diff
@@ -1,5 +1,46 @@
...
+
+* subexperiments: A term used to describe the unique circuit samples associated with a subcircuit. These circuits have had their `BaseQPDGate`\ s decomposed into local Qiskit gates and measurements. Subexperiments are the circuits sent to the backend for execution.
+
+* decompose: We try to honor the Qiskit notion of "decompose" in the documentation and API, which loosely means transforming a gate into a less-abstracted representation. *Ocassionally*, we may use the term "decompose" to refer to the act of inserting `BaseQPDGate` instances into quantum circuits as "decomposing" a gate or wire; however, we try to use terms like "partition" and "cut" when referring to this to avoid ambiguity with Qiskit language.
+
+Current limitations
+-------------------
+* QPD-based wire cutting available no sooner than CKT v0.3.0. The `cutqc <https://qiskit-extensions.github.io/circuit-knitting-toolbox/circuit_cutting/cutqc/index.htmlpackage>`_ package may be used for wire cutting in the meantime.
+* `PauliList` is the only supported observable format until no sooner than CKT v.0.3.0
```
garrison, line 26:
```suggestion
* ``PauliList`` is the only supported observable format until no sooner than CKT v0.3.0.
```
double backticks since this is rst, plus a small tweak

docs/circuit_cutting/explanation/index.rst
```diff
@@ -1,5 +1,46 @@
...
+
+* subexperiments: A term used to describe the unique circuit samples associated with a subcircuit. These circuits have had their `BaseQPDGate`\ s decomposed into local Qiskit gates and measurements. Subexperiments are the circuits sent to the backend for execution.
+
+* decompose: We try to honor the Qiskit notion of "decompose" in the documentation and API, which loosely means transforming a gate into a less-abstracted representation. *Ocassionally*, we may use the term "decompose" to refer to the act of inserting `BaseQPDGate` instances into quantum circuits as "decomposing" a gate or wire; however, we try to use terms like "partition" and "cut" when referring to this to avoid ambiguity with Qiskit language.
+
+Current limitations
+-------------------
+* QPD-based wire cutting available no sooner than CKT v0.3.0. The `cutqc <https://qiskit-extensions.github.io/circuit-knitting-toolbox/circuit_cutting/cutqc/index.htmlpackage>`_ package may be used for wire cutting in the meantime.
```
garrison, line 25:
```suggestion
* QPD-based wire cutting will be available no sooner than CKT v0.3.0. The `cutqc <https://qiskit-extensions.github.io/circuit-knitting-toolbox/circuit_cutting/cutqc/index.htmlpackage>`_ package may be used for wire cutting in the meantime.
```

docs/circuit_cutting/explanation/index.rst
```diff
@@ -1,5 +1,46 @@
...
+
+For more detailed information on this technique, refer to the paper, Error mitigation for short-depth quantum circuits [`4 <https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.119.180509>`_].
+
+Key terms
+-----------------
+* subcircuits: The set of circuits resulting from cutting gates in a `QuantumCircuit` and then separating the disconnected qubit subsets into smaller circuits. These circuits contain `SingleQubitQPDGate`\ s and will be used to instantiate each unique subexperiment.
```
garrison, line 17: need to either put double backticks around `QuantumCircuit` or put `:class:` immediately before the first backtick.  Same for the other classes that follow.

docs/circuit_cutting/explanation/index.rst
```diff
@@ -1,5 +1,46 @@
...
+
+For more detailed information on this technique, refer to the paper, Error mitigation for short-depth quantum circuits [`4 <https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.119.180509>`_].
```
garrison, line 13: I like to use arXiv links for these things, because the arXiv paper is available without a subscription, and it always links to the real paper.  An alternative could be to include both links.  Same comment where the reference is listed at the end of this document.

https://arxiv.org/abs/1612.02058

circuit_knitting_toolbox/circuit_cutting/cutting_decomposition.py
```diff
@@ -83,6 +99,151 @@ def partition_circuit_qubits(
...
+    Transform specified gates into :class:`TwoQubitQPDGate`\ s.
+
+    Args:
+        circuit: The circuit containing gates to be decomposed
+        gate_ids: The indices of the gates to decompose
+        inplace: Flag denoting whether to copy the input circuit before acting on it
+
+    Returns:
+        A copy of the input circuit with the specified gates replaced with :class:`TwoQubitGate`\ s
+        and a list of ``QPDBasis`` instances -- one for each decomposed gate.
```
garrison, line 120:
```suggestion
        and a list of :class:`QPDBasis` instances -- one for each decomposed gate.
```

docs/circuit_cutting/explanation/index.rst
```diff
@@ -1,5 +1,54 @@
...
+
+As mentioned above, the cost of any simulation based on quasiprobability distribution is an exponential sampling overhead. The overhead of a cut gate depends on which gate is cut; see the final appendix of [`1 <https://arxiv.org/abs/2205.00016>`__] for details.  Here, we will focus on the CNOT gate.  If no real-time classical communication is available between qubits of the cut gate or wire, cut CNOT gates incur a sampling overhead of O(:math:`9^n`), and wire cuts incur a sampling overhead of O(16\ :sup:`n`), where n is the total number of cuts. If real-time communication is available (i.e., if the hardware supports “dynamic circuits”), the sampling overhead for both CNOT gate and wire cuts may be reduced to O(:math:`4^n`) [`1 <https://arxiv.org/abs/2205.00016>`__] [`3 <https://arxiv.org/abs/2302.03366>`__]; however, support for circuit cutting with classical communication (LOCC) is not yet supported in CKT.
```
garrison, line 19: Tweaking my own contribution :)

```suggestion
As mentioned above, the cost of any simulation based on quasiprobability distribution is an exponential sampling overhead. The overhead of a cut gate depends on which gate is cut; see the final appendix of [`1 <https://arxiv.org/abs/2205.00016>`__] for details.  Here, we will focus on the CNOT gate.  If no real-time classical communication is available between qubits of the cut gate or wire, cut CNOT gates incur a sampling overhead of O(:math:`9^n`), and wire cuts incur a sampling overhead of O(:math:`16^n`), where :math:`n` is the total number of cuts. If real-time communication is available (i.e., if the hardware supports “dynamic circuits”), the sampling overhead for both CNOT gate and wire cuts may be reduced to O(:math:`4^n`) [`1 <https://arxiv.org/abs/2205.00016>`__] [`3 <https://arxiv.org/abs/2302.03366>`__]; however, support for circuit cutting with classical communication (LOCC) is not yet supported in CKT.
```

docs/circuit_cutting/tutorials/gate_cutting_to_reduce_circuit_depth.ipynb
```diff
@@ -0,0 +1,423 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "6c106db7",
+   "metadata": {},
+   "source": [
+    "### Replace distant gates with `TwoQubitQPDGate`s by specifying their indices\n",
+    "\n",
+    "`decompose_gates` will replace the gates in the specified indices with ``TwoQubitQPDGate``s and also return a list of ``QPDBasis`` instances -- one for each gate decomposition."
```
garrison, line 164:
```suggestion
    "`decompose_gates` will replace the gates in the specified indices with `TwoQubitQPDGate`s and also return a list of `QPDBasis` instances -- one for each gate decomposition."
```
since this is markdown

docs/circuit_cutting/tutorials/gate_cutting_to_reduce_circuit_depth.ipynb
```diff
@@ -0,0 +1,423 @@
...
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "from circuit_knitting_toolbox.circuit_cutting.qpd import decompose_qpd_instructions\n",
+    "\n",
+    "# Set some arbitrary bases to which the QPD gates should decompose\n",
+    "for idx in cut_indices:\n",
+    "    qpd_circuit[idx].operation.basis_id = 2\n",
```
garrison, line 265: Awesome, you got `basis_id = 2` to work :+1:

docs/circuit_cutting/tutorials/gate_cutting_to_reduce_circuit_depth.ipynb
```diff
@@ -0,0 +1,423 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "319c25d7",
+   "metadata": {},
+   "source": [
+    "### Reconstruct the simulated expectation values\n",
+    "\n",
+    "`reconstruct_expectation_values` expects `quasi_dists` and `coefficients` in the same format as returned from `execute_experiments`. `quasi_dists` is a 3D list of shape (num_unique_samples, num_partitions, num_commuting_observ_groups), and `coefficients` is a list with length equal to the number of unique samples. `subobservables` is the dictionary mapping qubit partition label to the associated subobservable(s). The decomposition of this circuit did not result in separation into subcircuits, so there will be only one partition represented in these outputs."
```
garrison, line 332:
```suggestion
    "`reconstruct_expectation_values` expects `quasi_dists` and `coefficients` in the same format as returned from `execute_experiments`. `quasi_dists` is a 3D list of shape (`num_unique_samples`, `num_partitions`, `num_commuting_observ_groups`), and `coefficients` is a list with length equal to the number of unique samples. `subobservables` is the dictionary mapping qubit partition label to the associated subobservable(s). The decomposition of this circuit did not result in separation into subcircuits, so there will be only one partition represented in these outputs."
```

docs/circuit_cutting/tutorials/gate_cutting_to_reduce_circuit_depth.ipynb
caleb-johnson: Yes, I was calling `QuantumCircuit.decompose()`, with no arguments, so it was sending `QPDMeasure` gates to the transpiler, which it doesnt know what to do with.

I had to use `decompose_qpd_instructions` to properly get everything to decompose down. I think it's cool to flash little pieces of the underlying functionality here and there in the tutorials and how-tos without distracting too much from the purpose of the doc

docs/circuit_cutting/tutorials/gate_cutting_to_reduce_circuit_width.ipynb
```diff
@@ -0,0 +1,370 @@
...
+    " - The coefficients for each subexperiment"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "f0032570",
+   "metadata": {},
+   "source": [
+    "### Reconstruct the simulated expectation values\n",
+    "`reconstruct_expectation_values` expects `quasi_dists` and `coefficients` in the same format as returned from `execute_experiments`. `quasi_dists` is a 3D list of shape (num_unique_samples, num_partitions, num_commuting_observ_groups), and `coefficients` is a list with length equal to the number of unique samples. `subobservables` is the dictionary mapping qubit partition label to the associated subobservable(s), as output from `decompose_problem` above."
```
garrison, line 279:
```suggestion
    "`reconstruct_expectation_values` expects `quasi_dists` and `coefficients` in the same format as returned from `execute_experiments`. `quasi_dists` is a 3D list of shape (`num_unique_samples`, `num_partitions`, `num_commuting_observ_groups`), and `coefficients` is a list with length equal to the number of unique samples. `subobservables` is the dictionary mapping qubit partition label to the associated subobservable(s), as output from `decompose_problem` above."
```

docs/circuit_cutting/tutorials/gate_cutting_to_reduce_circuit_depth.ipynb
caleb-johnson: @garrison I took this out at some point because the highlights get a bit dense in this paragraph, but I do agree these should be highlighted. That was my initial thought too

releasenotes/notes/gate-cutting-workflow-ebbedbdb1313b258.yaml
```diff
@@ -0,0 +1,43 @@
...
+    :class:`~circuit_knitting_toolbox.circuit_cutting.qpd.QPDBasis` and 
+    :class:`~circuit_knitting_toolbox.circuit_cutting.qpd.BaseQPDGate` for more information.
+
+    Addition of :mod:`~circuit_knitting_toolbox.circuit_cutting.cutting_decomposition`, :mod:`~circuit_knitting_toolbox.circuit_cutting.cutting_decomposition`,
+    and :mod:`~circuit_knitting_toolbox.circuit_cutting.cutting_decomposition` modules. These modules
+    provide several functions which allow for easy implementation of a gate cutting workflow, namely,
+    the `partition_problem`, `decompose_gates`, `execute_experiments`, and `reconstruct_expectation_value` functions.
+issues:
+  - |
+    Only support Pauli observables which have a trivial phase (1.0 + 0j). Full support for `SparsePauliOp` is expected
```
garrison, line 34: Keep in mind that this will be in a different section of the release notes than the above text, so it's important to establish context again.

docs/circuit_cutting/tutorials/gate_cutting_to_reduce_circuit_depth.ipynb
garrison: ha, I was just looking at it thinking the same thing before I saw this.  We can re-evaluate how this is presented in the future.

garrison (dismissed):
Thanks, this is a big one :partying_face:

releasenotes/notes/gate-cutting-workflow-ebbedbdb1313b258.yaml
```diff
@@ -0,0 +1,41 @@
...
+    :class:`~circuit_knitting_toolbox.circuit_cutting.qpd.QPDBasis` and 
+    :class:`~circuit_knitting_toolbox.circuit_cutting.qpd.BaseQPDGate` for more information.
+
+    Addition of :mod:`~circuit_knitting_toolbox.circuit_cutting.cutting_decomposition`, :mod:`~circuit_knitting_toolbox.circuit_cutting.cutting_decomposition`,
+    and :mod:`~circuit_knitting_toolbox.circuit_cutting.cutting_decomposition` modules. These modules
+    provide several functions which allow for easy implementation of a gate cutting workflow, namely,
+    the `partition_problem`, `decompose_gates`, `execute_experiments`, and `reconstruct_expectation_value` functions.
+issues:
+  - |
+    The :mod:`~circuit_knitting_toolbox.circuit_cutting` package only supports `PauliList` observables for calculating expectation values. Support for calculating
```
garrison, line 34:
```suggestion
    The :mod:`~circuit_knitting_toolbox.circuit_cutting` package only supports ``PauliList`` observables for calculating expectation values. Support for calculating
```

garrison (approved):

PR #156, opened 2023-05-12
Change `gate.operation` -> `instruction.operation`

garrison:
Currently, there are a few loops that use the variable `gate` for a `CircuitInstruction`.  This changes those to use the variable name `instruction` instead.  This removes the nonsensical `gate.operation` phrasing.  (`.operation` _might_ be a gate, but it is not in general.)


caleb-johnson (approved):
Thanks, I changed course on this after we started the repo. Been meaning to clean up these instances

caleb-johnson:
Ive been using `inst` instead of `instruction`. Up to you

garrison:
Yeah, actually I started with that, but then looked at `quantumcircuit.py`, and it typically uses `instruction`.  In a few places, it uses `instr` instead.

PR #162, opened 2023-05-12
Add `Reset` support to the simulator

garrison:
This adds support for reset operations to the simulator.  In doing so, it also removes the restriction that classical bits can only be written once.

This change is necessary to support the tests for wire cutting.


caleb-johnson (approved):
Awesome, one step closer to wire cutting 🕺

Issue #128, opened 2023-05-04
Publish to Zenodo


garrison:
- [ ] Obtain doi
- [ ] Add badge to README
- [ ] Update `CITATION.bib`

garrison:
I've now enabled Zenodo for this repository.  Instructions there say that the next step toward obtaining a doi is to make a release.  The doi should therefore happen automatically once we create the 0.2.0 release.

PR #165, opened 2023-05-12
Add bibtex note to Sphinx

garrison:
This partially addresses #128.  I also bumped the year in the citation.  We can/should update the author list soon, as well.  I built this locally, and it looks great.


caleb-johnson:
> This partially addresses #128. I also bumped the year in the citation. We can/should update the author list soon, as well. I built this locally, and it looks great.

Yes, I was looking at the authors last night. They need to be updated. Do we just strictly add to that (already long) list over time?

garrison:
> Yes, I was looking at the authors last night. They need to be updated. Do we just strictly add to that (already long) list over time?

I think so, at least for the time being.  I'll probably just update the authors in this same PR early next week.

caleb-johnson (approved):
Is this ready? Looks ok to me 👍

PR #174, opened 2023-05-13
Implement wire cutting as a two-qubit instruction

garrison:
- [x] wait for #153 to merge
- [x] add new `Move` instruction to Sphinx API References
- [x] write a tutorial notebook (**[preview here](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/wire-cutting/docs/circuit_cutting/tutorials/03_wire_cutting_via_move_instruction.ipynb)**)
- [x] remove the "limitation" from the cutting explanatory
- [x] remove wire cutting as a unique feature of cutQC
- [x] release note


caleb-johnson:
We'll need to remove the "limitation" from the cutting explanatory when this goes in.

https://qiskit-extensions.github.io/circuit-knitting-toolbox/circuit_cutting/explanation/index.html

circuit_knitting/cutting/instructions/move.py
```diff
@@ -0,0 +1,50 @@
...
+    second incoming qubit must share no entanglement with the remainder of the
+    system.  If this qubit *is* entangled, then performing the reset operation
+    will in turn implement a quantum channel on the other qubit(s) with which
+    it is entangled, resulting in the partial collapse of those qubits.
+
+    Generally speaking, the most straightforward way to ensure that the second
+    qubit is not entangled is to prepare it in one of the following ways before
+    implementing the :class:`Move`:
+
+    - No operations since initialization.
```
caleb-johnson, line 33: This is a great writeup, but I find these bullets, specifically, a bit confusing.

Maybe you can explicitly say in the first bullet "The qubit should contain no operations since circuit initialization" or something. I know it's more verbose, but it'll make it 100% clear we're talking about preparing this 2nd qubit.

I think the rest will be more clear if you add to the first point, but I am a bit confused about both `Measure` and `Reset` being the "preceding" operation. I would think only one could directly precede the `Move`. I would just try to make this a bit more clear.

I'm a bit confused on the last point too. This seems like it is an either/or with the other three items, but it's lumped in like all 4 are strong suggestions for preparing a `Move_2ndQ`.

Maybe drawing an example circuit in the docstring is appropriate here?

caleb-johnson (commented):
This looks great.

One question: I understand that the `Move` instruction decomposes to a swap and a reset when "decomposed" by Qiskit. What I don't understand is when in our workflow that SWAP happens or when we add the SingleQubitQPDGates containing the wire cut QPDBasis. Going to keep studying it but maybe you can say something about how this happens in our current workflow

caleb-johnson:
> This looks great.
> 
> One question: I understand that the `Move` instruction decomposes to a swap and a reset when "decomposed" by Qiskit. What I don't understand is when in our workflow that SWAP happens or when we add the SingleQubitQPDGates containing the wire cut QPDBasis. Going to keep studying it but maybe you can say something about how this happens in our current workflow

After looking at the tutorial, I think #326 is what I was missing. Looks good, modulo the a small comment or two

circuit_knitting/cutting/instructions/move.py
garrison: I tried to improve this in 88f98d3cf6dd5d1e8a2a93a503d09fb9cbf68d61.  Rather than list all of the possibilities (which, actually, my previous list didn't even do exhaustively), I tried to focus on the two main ways in which you can satisfy the condition: use a fresh qubit as destination qubit OR use a qubit that was previously the source of a different move.  The two move instructions in the tutorial correspond to these cases, respectively.

garrison: > Maybe drawing an example circuit or two in the docstring is appropriate here?

I'm going to look into how to draw circuits within docstrings, too.

garrison: I pushed some further improvements in c716d9854788649b691db61cc7326341f8e9c6ef.

circuit_knitting/cutting/instructions/move.py
```diff
@@ -0,0 +1,82 @@
...
+       :include-source:
+
+       import numpy as np
+       from qiskit.circuit import QuantumCircuit, CircuitInstruction
+       from circuit_knitting.cutting.instructions import Move
+
+       qc = QuantumCircuit(4)
+       qc.ryy(np.pi / 4, 0, 1)
+       qc.rx(np.pi / 4, 3)
+       qc.append(CircuitInstruction(Move(), [qc.qubits[1], qc.qubits[2]]))
```
garrison, line 60: @SaashaJoshi realized, in #326, that there's no need to explicitly create a `CircuitInstruction` here.  I'm going to update this PR accordingly.

garrison: Fixed in 56ea75821d4bc568a44294d4c4d3b0fa44ee2b09.

caleb-johnson (approved):
This looks really nice! LGTM 👍

PR #190, opened 2023-05-19
Use exact weights in `generate_qpd_samples` when it's possible for _all_ of them

garrison:
This changes `generate_qpd_samples` to use exact weights when it's possible for _all_ of them, i.e., it's either all or nothing, for now at least.  I am working on a subsequent PR that will allow a mixture: the weights above the `1 / num_samples` threshold will be evaluated exactly, while everything else will be sampled.   However, that's a bit more complex, and this will suffice for some of the basic use cases (e.g., cutting CNOTs only).


caleb-johnson (approved):
awesome, looks gtm 👍

PR #193, opened 2023-05-22
Don't add empty cregs to circuits

caleb-johnson:
This PR prevents qpd._decompose_qpd_measurements from adding empty QPD registers. This came up while creating unit tests for #153 . Removing the empty registers makes it more clear what is happening under the hood and makes it more straightforward to test against.


caleb-johnson:
This PR makes less sense outside the context of the cutting tests. I'm just going to roll this change into #153

PR #194, opened 2023-05-22
Allow `num_samples` to be a float

garrison:
This change allows `num_samples` to be a float.  This really only has one immediate impact: one can pass `math.inf` as `num_samples` and get exact weights (no sampled ones) no matter what.  While ordinarily the weights will all add up to `num_samples,` this is not possible when `num_samples == math.inf`, so the weights instead add up to 1 in this special case.

Otherwise, this is the first step in preparation for my PR that will allow both sampled and exact values to be returned together -- so it's not all or nothing (#190).

Why does a float as `num_samples` make sense in general?  There is this special threshold given by `1 / num_samples`, and this change allows tweaking this threshold arbitrarily.

### DIFF
--- a/circuit_knitting_toolbox/circuit_cutting/qpd/qpd.py
+++ b/circuit_knitting_toolbox/circuit_cutting/qpd/qpd.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 from collections.abc import Sequence, Callable
 from enum import Enum
 import itertools
+import math
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


garrison:
If this makes total sense to you, Caleb, we can merge it now.  But I added it to the 0.3.0 milestone because that's the real timeframe in which I think we should evaluate this.

caleb-johnson (approved):
Hey this looks good to me 👍

PR #198, opened 2023-05-25
Improve the API reference docs within the utils directory

garrison:
This improves the API docs within the utils, by showing each utility module separately on the utils page.  I've also enabled Sphinx parallel build to save time (45 seconds, down from 75 seconds before).

### DIFF
--- a/circuit_knitting_toolbox/utils/__init__.py
+++ b/circuit_knitting_toolbox/utils/__init__.py
@@ -12,18 +12,14 @@
 """
 Utility functions.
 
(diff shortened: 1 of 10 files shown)


garrison:
Ugh, pydocstyle doesn't like these header lines, even though that's what they do in Terra.

garrison:
I suppose I will need to move all the headers to `__init__.py`.

caleb-johnson (approved):
Great, thanks 💯

PR #201, opened 2023-05-26
Rephrase qpd explanation

garrison:
This is much longer now, perhaps somewhat redundant, but it's more precise.


caleb-johnson (approved):
Looks great, thanks. Merge when ready

PR #204, opened 2023-05-26
Reorganize optional dependencies, and use randomly in notebook tests

garrison:
I noticed that our notebook tests don't use `pytest-randomly`, so any given run is not reproducible.  While addressing this, I also re-organized the "optional" dependencies a bit.


caleb-johnson (approved):
LGTM thanks 👍

PR #206, opened 2023-05-27
Remove `ThreadPool` in qpd cutting evaluation code

garrison:
This seems to have been leading to deadlock, perhaps because qiskit-ibm-runtime may not be thread safe.

I've adjusted it to submit all the jobs, then collect the results, all on the same thread.

It's a bit obscure use of generators, but it works, and we can refactor later.


garrison:
Closing in favor of #207

PR #210, opened 2023-05-27
Render QPD API in sphinx api docs

caleb-johnson:
There are QPD classes referenced in the cutting documentation, so the QPD package should be searchable in the API, even if we don't otherwise advertise it or create notebook tutorials centered around it.

- Added QPD section to circuit cutting section of API docs 
- Added CutQC section to circuit cutting section of API docs
 - Added a [no_inherited_members template](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/render-qpd-api/docs/_templates/autosummary/class_no_inherited_members.rst) for preventing rendering of inherited members and applied it to the classes in the [QPD](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/render-qpd-api/circuit_knitting_toolbox/circuit_cutting/__init__.py) section of the circuit cutting init. This is a copy of the [default class.rst](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/ext/autosummary/templates/autosummary/class.rst) in the Sphinx repo with a few modifications to suit our needs:
     - It prevents inherited members from rendering. We do not want `qiskit.circuit.Instruction` API to mix with our QPDGate API, even though `BaseQPDGate` extends `Instruction`. I won't list the reasons here, but we can discuss more.
     - It addds inheritance classes to the class pages in the API docs. Since we're not printing inheritance members, we will at least print the classes we inherit from
     - It prevents duplicating the init signature rendering by removing the automember call to init contained in the default class.rst
 - Removed unused autosummary docstrings from `qpd` and `qpd.instructions` packages. These autosummary docstrings are captured in the `circuit_cutting` package. Can revisit API docs hierarchy at a later time. As the modules grow, we will likely want to allow the API docs hierarchy to grow deeper, and we will rewrite these autosummary docstrings at that time.


caleb-johnson:
I wish our classes didnt render their parent class methods and attributes in the API (e.g. BaseQPDGate renders `qiskit.circuit.Instruction` methods and attributes)

caleb-johnson:
> I wish our classes didnt render their parent class methods and attributes in the API (e.g. BaseQPDGate renders `qiskit.circuit.Instruction` methods and attributes)

Accomplished with `:template: autosummary/class_no_inherited_members.rst`. I do like this much better

docs/_templates/autosummary/class_no_inherited_members.rst
```diff
@@ -0,0 +1,33 @@
...
+
+   {% block methods %}
+   .. automethod:: __init__
+
+   {% if methods %}
+   .. rubric:: {{ _('Methods') }}
+
+   .. autosummary::
+   {% for item in methods %}
+   {%- if item not in inherited_members %}
```
caleb-johnson, line 15: This line, line 28, and their associated if-close statements are the only things added to this file from the [default class.rst](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/ext/autosummary/templates/autosummary/class.rst) in Sphinx.

caleb-johnson: Resolving. Left a comment saying same thing in the file

circuit_knitting_toolbox/circuit_cutting/__init__.py
```diff
@@ -14,6 +14,51 @@
...
+    decompose_gates
+    execute_experiments
+    reconstruct_expectation_values
+
+.. autosummary::
+    :toctree: ../stubs/
+    :nosignatures:
+    :template: autosummary/class_no_inherited_members.rst
+
+    PartitionedCuttingProblem
```
caleb-johnson, line 35: These named tuples need to be included in the API docs because they are referenced as return types in our cutting functions

docs/_templates/autosummary/class_no_inherited_members.rst
garrison, line 1: Did you obtain this by modifying the default class template?  It might be worth saying this explicitly, so we know what to do if we ever run into issues, e.g., if there are changes to the default template in a future version of Sphinx.

garrison: And yes, I see the comment; I'm just trying to get you to change "This is very similar to" into something a little bit more precise.  Something like copied from file xxx in v.x of Sphinx, with the following modifications.

garrison (approved):
Thanks especially for figuring out this no-inherited-members thing.

Issue #216, opened 2023-05-30
Prepend many Sphinx class/func/mod cross references with a dot


garrison:
In many places, we say something like ``` :class:`Something` ```, but it would be even better to say ``` :class:`.Something` ```, as this would allow Sphinx to resolve the cross references in our documentation, which should then make these references clickable.

From [the docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html):

> if the name is prefixed with a dot, and no exact match is found, the target is taken as a suffix and all object names with that suffix are searched. For example, :py:meth:`.TarFile.close` references the tarfile.TarFile.close() function, even if the current module is not tarfile. Since this can get ambiguous, if there is more than one possible match, you will get a warning from Sphinx.
>
> Note that you can combine the ~ and . prefixes: :py:meth:`~.TarFile.close` will reference the tarfile.TarFile.close() method, but the visible link caption will only be close().

PR #217, opened 2023-05-30
Tweaks to gate cutting workflow release note

garrison:
- use proper sphinx cross references for tutorials & explanation, to prevent bitrot
- fix a few typos
- prefix function xrefs with dots (see also #216)


caleb-johnson (approved):
Nice, thanks 👍

PR #219, opened 2023-05-30
Ensure that each sampler passed to `samplers` is unique

garrison:
In the future, we might want the code to behave differently based on whether samplers are identical or merely equivalent.  For instance, perhaps we will want identical samplers to mean that the circuit is evaluated all at once on the hardware, while two equivalent samplers could mean the subcircuits are passed to the backend twice independently (see #220 for details).  For now, since we might want to do this in the future, I believe it makes sense to explicitly disallow it until we figure this out.

I also fixed an unrelated typo in an exception message.

### DIFF
--- a/circuit_knitting_toolbox/circuit_cutting/cutting_evaluation.py
+++ b/circuit_knitting_toolbox/circuit_cutting/cutting_evaluation.py
@@ -94,12 +94,24 @@ def execute_experiments(
     if isinstance(circuits, dict):
         if circuits.keys() != subobservables.keys():
             raise ValueError(
-                "The keys for the circuits and observabes dicts should be equivalent."
+                "The keys for the circuits and observables dicts should be equivalent."
             )
         if isinstance(samplers, dict) and circuits.keys() != samplers.keys():
             raise ValueError(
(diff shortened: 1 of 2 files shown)


circuit_knitting_toolbox/circuit_cutting/cutting_evaluation.py
```diff
@@ -94,13 +94,25 @@ def execute_experiments(
...
             raise ValueError(
-                "The keys for the circuits and observabes dicts should be equivalent."
+                "The keys for the circuits and observables dicts should be equivalent."
             )
         if isinstance(samplers, dict) and circuits.keys() != samplers.keys():
             raise ValueError(
                 "The keys for the circuits and samplers dicts should be equivalent."
             )
 
+    if isinstance(samplers, dict):
```
caleb-johnson, line 104: Looks great. Could we add this to the docstring above?

garrison: Done

caleb-johnson (approved):
Nice, thanks 👍

PR #222, opened 2023-05-30
Fix the workflow if an empty circuit is provided

### LINKED ISSUE #218: Workflow fails on empty circuit
I'd expect this to work, and it _does_ if I make `qc` prepare a Bell state, but it doesn't if I leave `qc` empty.

```python
from qiskit import QuantumCircuit
from qiskit_aer.primitives import Sampler
from circuit_knitting_toolbox.circuit_cutting import partition_problem, execute_experiments

qc = QuantumCircuit(2)
subcircuits, _, subobservables = partition_problem(
    circuit=qc, partition_labels="AB", observables=PauliList(["XX"])
)
execute_experiments(
    subcircuits,
    subobservables,
    num_samples=10,
    samplers=Sampler(),
)
```

```
ValueError: The keys for the circuits and observables dicts should be equivalent.
```

garrison:
Closes #218.  I started by modifying `_separate_instructions_by_partition`, but then I encountered a new error, which led me to modify `_circuit_from_instructions` as well.

The only thing I was a little bit unsure about was whether `qc.qubits[j]` would indeed give me the `j`th qubit, but it seems to work: all tests pass.

As a bonus, each subsystem now has a _single_ quantum register which contains all qubits relevant to it.

### DIFF
--- a/circuit_knitting_toolbox/utils/transforms.py
+++ b/circuit_knitting_toolbox/utils/transforms.py
@@ -34,7 +34,9 @@ from qiskit.circuit import (
     QuantumRegister,
     ClassicalRegister,
     Barrier,
+    Qubit,
 )
+from .iteration import unique_by_eq
 
 
 class SeparatedCircuits(NamedTuple):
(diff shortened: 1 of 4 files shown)


caleb-johnson (approved):
Great, thanks 👍

PR #227, opened 2023-05-31
Create issue templates for CKT

caleb-johnson:


garrison:
Hmm, it's not self evident to me what the difference between a feature and enhancement is.  I could imagine myself clicking on both to see how the templates are different, and I might have an easy time deciding in some case, but not all.  Couldn't they just all be considered enhancements?

caleb-johnson:
> Hmm, it's not self evident to me what the difference between a feature and enhancement is. I could imagine myself clicking on both to see how the templates are different, and I might have an easy time deciding in some case, but not all. Couldn't they just all be considered enhancements?

I ported this from the template repo and did a few minor updates. I was going to put a little more effort into it and fill out the PR before pinging you.

I agree, those can be combined. I also hadn't decided yet whether the feature request template should be fleshed out more. Also just wanted to get your opinion on the bug report template as a whole, which is the most important imo. You may have been the one to write this in template repo, I can't remember

caleb-johnson:
This has gone stale, closing until ready to revisit

PR #232, opened 2023-06-01
Add backwards compatibility tests of 0.2 circuit cutting workflows

garrison:
This will make sure we don't _accidentally_ break the workflows in the current tutorial notebooks as we tweak the API going forward.  We might deliberately break the original workflow, but this should be an intentional choice, not something we do by accident.


garrison:
I inspected the CI logs, and, as expected, all but four tests get run for coverage, and all tests are run for the other tests.

caleb-johnson (approved):
Great idea, thanks. Love the pytest magic

caleb-johnson (approved):
Great idea, thanks. Love the pytest magic

PR #233, opened 2023-06-01
Rename the tutorials to order by number & enable redirects to prevent link rot

garrison:
My goal here is for the icons on the [cutting tutorials](https://qiskit-extensions.github.io/circuit-knitting-toolbox/circuit_cutting/tutorials/index.html) page to be listed in the order that they are meant to be consumed (i.e., to match the order in the list on that page).  The "gallery" is sorted by filename, so the way to do this is to rename the files to have a number, so that lexicographic sorting will put them in the correct order.

I've also enabled and configured [sphinx-reredirects](https://gitlab.com/documatt/sphinx-reredirects) so that redirects will be placed at the old URLs, in order to prevent link-rot, because [cool URIs don't change](https://www.w3.org/Provider/Style/URI).  I've built the docs locally to make sure the redirects work.


caleb-johnson (approved):
Looks good thanks 👍

Issue #221, opened 2023-05-30
Shorten top-level module names


garrison:
We've talked before about shortening some of the import names.  I propose changing

```python
from circuit_knitting_toolbox.entanglement_forging import EntanglementForgingAnsatz
from circuit_knitting_toolbox.circuit_cutting import partition_circuit_qubits
```

to

```python
from circuit_knitting.forging import EntanglementForgingAnsatz
from circuit_knitting.cutting import partition_circuit_qubits
```

#### Rationale

- `circuit_knitting` is descriptive enough.  `_toolbox` does not need to be mentioned in every import location.  Previously, we've considered `ckt`, but this is non-descriptive and could potentially clash with another package (the three letter namespace is typically very overloaded)
- `circuit_knitting.circuit_cutting` is redundant.  `circuit_knitting.cutting` is less of a mouthful
- I'm more torn between `circuit_knitting.entanglement_forging` and `circuit_knitting.forging`; however, if somebody is using this module, they probably have a good idea of what "forging" is about.

#### Other notes

Let's keep the documentation directories the way they are rather than replace `circuit_cutting` and `entanglement_forging` with shorter names.  This will
1. prevent links from unnecessarily breaking
2. provide more descriptive names to people navigating the documentation, e.g., the tutorials

garrison:
If we want to do this, I think we will need to switch from flit to hatch/hatchling.  From what I can tell, flit only supports a single top-level module name, and this is a deliberate design decision.  We will need two for a bit, since we want to leave `circuit_knitting_toolbox` as a deprecated import location (so as not to break existing code).

PR #234, opened 2023-06-01
Switch from flit to hatchling build system

garrison:
This switches from flit_core to [hatchling](https://pypi.org/project/hatchling/) for the build system.  Likewise, it uses [hatch](https://github.com/pypa/hatch) rather than flit in the `release.yml` workflow.

The primary goal here is to enable progress on
- #221

I have not tested the release workflow against the test pypi instance, but everything _looks_ right to me.  If it fails when we do the 0.3.0 release, I am happy to debug.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,13 +31,17 @@ jobs:
         uses: actions/checkout@v3
         with:
           ref: ${{ github.ref_name }}
-      - name: Install flit
+      - name: Install hatch
         run: |
           python -m pip install --upgrade pip
-          pip install flit
+          pip install hatch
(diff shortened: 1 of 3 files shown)


caleb-johnson (approved):
https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/issues/235

Sounds great, I opened up an issue to test against the pypi test server ahead of 0.3.0 release day. I'll try to play with this in the next couple weeks

garrison:
> I opened up an issue to test against the pypi test server ahead of 0.3.0 release day. I'll try to play with this in the next couple weeks

Thanks, that'd be awesome :)

PR #238, opened 2023-06-02
Make coverage fail if under 90%

garrison:


caleb-johnson (approved):
Great! How has our coverage of the older code crept up above 90? Wasn't it high 80s before?

garrison:
> Great! How has our coverage of the older code crept up above 90? Wasn't it high 80s before?

Well, it's really an _overall_ coverage percentage, and as we have more code at 100% coverage, the overall fraction goes up too :)

PR #242, opened 2023-06-04
QPD gates generated through cutting should have "cut_" label prefix

### LINKED ISSUE #241: partition_circuit_qubits should produce gates called cut_{gatename}
For symmetry with decompose_gates, and since it ended up being a cutting function, partition_circuit_qubits should produce qpd gates with the prefix "cut" rather than "qpd".

caleb-johnson:
Resolves #241 

QPD gates generated through the cutting API should have the prefix "cut" in their label, rather than "qpd".

We shifted `partition_circuit_qubits` from the QPD repo internally to the cutting package externally, but did not update this label. We have also started to tighten up usage of the term decompose in our API and documentation to be more in-line with Qiskit's notion, and I think referring to these gates as "cut" when they have been "partitioned" and possibly "separated" is best. If we can avoid the word "decompose" or "decomposition" until referring to the actual creation of subexperiments, I think it will make things more clear. 

Part of me wishes we went with the original "quasi-probability representation (QPR)" or Christophe Piveteau's terminology, "quasi-probability simulation", but that's another discussion :)

This is a breaking change, but I think it is OK to just document this and get it through ASAP.

- [x] Make release note


releasenotes/notes/update-qpdgate-labels-612e7d355ae1c1a3.yaml
```diff
@@ -0,0 +1,8 @@
+---
+upgrades:
```
caleb-johnson, line 2: Kinda weird to put a breaking change in "upgrades", but it is a code quality upgrade, and it is in line with [reno docs](https://docs.openstack.org/reno/latest/user/usage.html).

```
deprecations
A list of features, APIs, configuration options to be deprecated in the release. Deprecations should not be used for something that is removed in the release, use upgrade section instead. Deprecation should allow time for users to make necessary changes for the removal to happen in a future release.
```

PR #243, opened 2023-06-04
Deprecate decompose_gates

### LINKED ISSUE #239: Rename decompose_gates to cut_gates
We want to honor Qiskit's definition of "decompose", which loosely means to transform a quantum gate to a less abstracted representation, with respect to the compiler.

In that spirit, the "decompose_gates" function should become "cut_gates", as we are replacing non-local gates with a different, *more* abstract representation. Additionally, "decompose_gates" produces gates with labels of the form "cut_{gate_name}", so this naming is more natural in that way too.

Resolution of this issue includes deprecating the "decompose_gates" function name, implementing an identical "cut_gates" interface, and updating all documentation accordingly.

caleb-johnson:
Resolves #239 

This PR deprecates `decompose_gates` in favor of `cut_gates`. This change is solely to remove ambiguity between Qiskit's notion of "decompose" and ours.


circuit_knitting_toolbox/circuit_cutting/__init__.py
```diff
@@ -23,7 +23,7 @@
 
     partition_circuit_qubits
     partition_problem
-    decompose_gates
```
caleb-johnson, line 26: This actually needs to stay in the API docs, and I need to update the docstring to say "deprecated"

circuit_knitting/utils/transforms.py
garrison, line 1: If you run `git merge main` on this branch locally (after making sure `main` is up to date), it should remove this file from the diff here.

caleb-johnson: I tried to `git checkout main [filepath]` and it did a no-op update. Thanks

circuit_knitting/cutting/cutting_decomposition.py
```diff
@@ -104,8 +105,29 @@ def partition_circuit_qubits(
     return circuit
 
 
+@deprecate_func(
+    since="0.3",
+    package_name="circuit-knitting-toolbox",
+    removal_timeline="no earlier than v0.4.0",
+    additional_msg=(
+        "Instead, use ``circuit_knitting.cutting.cut_gates`` "
```
caleb-johnson, line 113: @garrison do we need back quotes here? Is this strictly a stdio printout, or is this used by sphinx?

garrison: I only included double backquotes because that's what terra did in the opflow deprecation message.  If it weren't for that, I probably would have used single backquotes, but nothing is required.

caleb-johnson: I actually thought I wrote this out of habit, but thank you for your honesty hah :)

PR #244, opened 2023-06-06
Rename top-level module to `circuit_knitting`

### LINKED ISSUE #221: Shorten top-level module names

garrison:
This performs the renames porposed in #221 and thus closes #221.

In order to demonstrate that the deprecated import locations work, I have not updated the imports in the tests or notebooks yet.  I'll do that once CI passes.

I also made it so docs will no longer deploy for pushes to `main`, since `main` will now be ahead of the current stable release.


circuit_knitting/__init__.py
garrison, line 1: This PR needs a release note. I will add one in the morning.

pyproject.toml
```diff
@@ -93,12 +93,18 @@ notebook-dependencies = [
 "Documentation" = "https://qiskit-extensions.github.io/circuit-knitting-toolbox/"
 "Repository" = "https://github.com/Qiskit-Extensions/circuit-knitting-toolbox"
 
+[tool.hatch.build.targets.wheel]
+only-include = [
+    "circuit_knitting",
```
caleb-johnson, line 98: So users will have access to two namespaces, and one of the namespaces will throw a bunch of warnings?

caleb-johnson (approved):
This looks really great, nice job. I had a couple questions out of interest but lgtm

circuit_knitting_toolbox/circuit_cutting/__init__.py
```diff
@@ -90,3 +39,35 @@
     "PartitionedCuttingProblem",
     "CuttingExperimentResults",
 ]
+
+sys.modules["circuit_knitting_toolbox.circuit_cutting.qpd"] = qpd
```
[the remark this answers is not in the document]
garrison: It tells the module system that if somebody tries to import a module called `circuit_knitting_toolbox.circuit_cutting.qpd`, then instead use this other thing that's already been loaded, which in this case is `circuit_knitting.cutting.qpd`.

This is explained at bit at https://stackoverflow.com/a/72244240/1558890.  (I tried the lazy method mentioned there too, but it had more issues than was worth.)

pyproject.toml
garrison: Exactly.

garrison:
> One more thing, should we rename the docs directories to `cutting` and `forging` ?

This was an intentional choice, but we could re-evaluate.  Quoting from #221:

> Let's keep the documentation directories the way they are rather than replace circuit_cutting and entanglement_forging with shorter names. This will
>
> - prevent links from unnecessarily breaking
> - provide more descriptive names to people navigating the documentation, e.g., the tutorials

caleb-johnson:
> > One more thing, should we rename the docs directories to `cutting` and `forging` ?
> 
> This was an intentional choice, but we could re-evaluate. Quoting from #221:
> 
> > Let's keep the documentation directories the way they are rather than replace circuit_cutting and entanglement_forging with shorter names. This will
> > 
> > * prevent links from unnecessarily breaking
> > * provide more descriptive names to people navigating the documentation, e.g., the tutorials

I agree with both. I don't think this is a source of confusion

releasenotes/notes/rename-top-level-import-bad074e7a8f348c9.yaml
```diff
@@ -0,0 +1,14 @@
+---
+upgrade:
+  - |
+    The top-level name for imports has been renamed from
```
caleb-johnson, line 4: Aren't deprecations normally replaced with a new target or a new way to migrate? How is this different? Isn't it a deprecation?

caleb-johnson (approved):

PR #245, opened 2023-06-08
Fix mypy under rustworkx 0.13.0

garrison:
This fixes mypy when run with the most recent rustworkx release, 0.13.0.  Without this, mypy errors with

```
circuit_knitting/utils/transforms.py:30: error: Module "rustworkx" has no attribute "connected_components"; maybe "BiconnectedComponents"?  [attr-defined]
circuit_knitting/utils/transforms.py:102: error: Need type annotation for "graph"  [var-annotated]
Found 2 errors in 1 file (checked 36 source files)
```

It's not clear to me, though, why these annotations are necessary.  I expect there is a better solution.

### DIFF
--- a/circuit_knitting/utils/transforms.py
+++ b/circuit_knitting/utils/transforms.py
@@ -27,7 +27,7 @@ from collections import defaultdict
 from collections.abc import Sequence, Iterable, Hashable, MutableMapping
 from typing import NamedTuple
 
-from rustworkx import PyGraph, connected_components
+from rustworkx import PyGraph, connected_components  # type: ignore[attr-defined]
 from qiskit.circuit import (
     QuantumCircuit,
     CircuitInstruction,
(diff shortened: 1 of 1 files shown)


garrison:
After opening this, I learned that rustworkx 0.13.0 introduced type annotations, but they are currently incomplete.  Since this fixes CI, I am in favor of merging it in its current form.

circuit_knitting/utils/transforms.py
```diff
@@ -27,7 +27,7 @@
 from collections.abc import Sequence, Iterable, Hashable, MutableMapping
 from typing import NamedTuple
 
-from rustworkx import PyGraph, connected_components
+from rustworkx import PyGraph, connected_components  # type: ignore[attr-defined]
```
caleb-johnson, line 30: I like this. I wonder how we can know when to remove it? Should we make an issue to keep up with this?

caleb-johnson: ok sounds good!

caleb-johnson (approved):
Looks good, thanks! We'll just check in on this when rustworkx has fully fleshed out their type hints

PR #246, opened 2023-06-11
Enable tests on Windows

garrison:
I realized that the entirety of circuit cutting works on Windows, so we might as well advertise native Windows support and test on it too.

Specific changes in this PR:
- skip tests that require pyscf if it is not available
- add Windows workflow to Actions matrix
- add Windows to README platform badge


caleb-johnson (approved):
Sweet!

PR #249, opened 2023-06-11
Add note on samping overhead to cutting tutorials

### LINKED ISSUE #225: Explain sampling overhead in tutorials
Each circuit cutting tutorial has a section for sampling overhead. Enter a sentence or two explaining its significance

caleb-johnson:
Resolves #225


docs/circuit_cutting/tutorials/01_gate_cutting_to_reduce_circuit_width.ipynb
garrison, line 1: "two" -> "three" in the last sentence in this file.  The expression is right, but the word is wrong.

garrison (commented):
> The sampling overhead is the required number of samples needed to get an expectation value of a target observable within some error, $\epsilon$.

How about:

The sampling overhead is the factor by which the number of samples must increase for the quasiprobability decomposition to result in the same amount of error as one would get by sampling the original circuit.

caleb-johnson:
> How about:
> 

Hmm your comment doesnt say `Outdated`, but I made this change

docs/circuit_cutting/tutorials/01_gate_cutting_to_reduce_circuit_width.ipynb
garrison, line 1: Remaining suggestions:
- "amount of error as one would get" -> "amount of error, `$\epsilon$`, as one would get" (sorry, this is my bad -- I didn't realize with my suggestion before that you actually use epsilon in the expressions that follow)
- `k` -> `$k$`
- I would suggest just putting single `$`s around the entire expression, parentheses and all, instead of all the interior `$`s to have the parentheses and / be rendered in the text font.  I'm guessing there is a reason you did this though -- did it mess with spacing in a weird way?

(Same comments apply to the other file, too.)

caleb-johnson: I can't say for certain. I was using jupyter to hack around with it until it looked right. Let me see if I can tidy them up some

garrison (approved):
Thanks, this looks good.  One final thought: I think the word "roughly" may be unnecessary, especially since it is followed by big-O, which is really about asymptotic behavior.  I did not even notice this the first five times I read it, so it is a pretty minor point.

caleb-johnson:
> Thanks, this looks good. One final thought: I think the word "roughly" may be unnecessary, especially since it is followed by big-O, which is really about asymptotic behavior. I did not even notice this the first five times I read it, so it is a pretty minor point.

I took this from the paper and wondered why it was rough. Is it because different gates have different base factors, or is it because there is something inexact going on with the chosen epsilon? I'm going to remove it since I'm not certain either

PR #263, opened 2023-06-16
Export `qpdbasis_from_gate` from `qpd` module

garrison:
`qpdbasis_from_gate` is listed in `__all__`, but it currently doesn't get exported because it hasn't been imported.  Actually, this manifests in the following error:

```
>>> from circuit_knitting.cutting.qpd import *
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'circuit_knitting.cutting.qpd' has no attribute 'qpdbasis_from_gate'
```

(I'm kind of disappointed that ruff doesn't pick this up for us.)


caleb-johnson (approved):
Nice catch, thanks

PR #267, opened 2023-06-17
Improve `separate_circuit` docstring

garrison:
I've also added `circuit_knitting/` to the paths passed to pytest, as this is required for doctests to be discovered.

### DIFF
--- a/circuit_knitting/cutting/cutqc/mip_model.py
+++ b/circuit_knitting/cutting/cutqc/mip_model.py
@@ -17,15 +17,6 @@ from typing import Sequence, Any
 
 import numpy as np
 
-try:
-    from docplex.mp.model import Model
-    from docplex.mp.utils import DOcplexException
-except ModuleNotFoundError as ex:  # pragma: no cover
-    raise ModuleNotFoundError(
-        "DOcplex is not installed.  For automatic cut finding to work, both "
-        "DOcplex and cplex must be available."
-    ) from ex
-
 
 class MIPModel(object):
     """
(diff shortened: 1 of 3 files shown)


circuit_knitting/utils/transforms.py
```diff
@@ -52,15 +64,41 @@ def separate_circuit(
...
 
+    If ``partition_labels`` is provided, then the circuit will be separated
+    according to those labels.  If it is ``None``, then the circuit will be
+    fully separated into its disconnected components, each of which will be
+    labeled with consecutive integers starting with 0.
+
+    >>> qc = QuantumCircuit(4)
+    >>> _ = qc.x(0)
+    >>> _ = qc.cx(1, 2)
+    >>> _ = qc.h(3)
```
garrison, line 75: Without the `_ = `, one gets extra output from the non-`None` expressions:

```
>>> from qiskit import QuantumCircuit
>>> qc = QuantumCircuit(4)
>>> qc.x(0)
<qiskit.circuit.instructionset.InstructionSet object at 0x7f710e9921c0>
>>> qc.cx(1, 2)
<qiskit.circuit.instructionset.InstructionSet object at 0x7f70da3873a0>
>>> qc.h(3)
<qiskit.circuit.instructionset.InstructionSet object at 0x7f70da3872b0>
```

and doctest fails unless these are also added here.  Of course, the memory address is not predictable, so there is no way to allow such a thing to be output and still have the tests pass.

circuit_knitting/cutting/cutqc/mip_model.py
garrison, line 1: This file had to be modified for doctest discovery to work on Python 3.11 (i.e., when DOcplex is not present).

circuit_knitting/utils/transforms.py
```diff
@@ -40,7 +45,14 @@
 
 
 class SeparatedCircuits(NamedTuple):
-    """Named tuple for result of :class:`separate_circuit`."""
+    """Named tuple for result of :func:`separate_circuit`.
+
+    ``subcircuits`` is a dict of circuits, keyed by each partition label.
+    ``qubit_map`` is a list of length number of qubits in the original circuit.
```
caleb-johnson, line 51:
```suggestion
    ``qubit_map`` is a list with length equal to the number of qubits in the original circuit.
```

caleb-johnson (approved):
Awesome, much needed, thanks. Thanks for adding it to api docs as well

circuit_knitting/utils/transforms.py
garrison: Oh man, thank you.  I rewrote this countless times and was never quite satisfied.

PR #230, opened 2023-06-01
Suppress warnings from Qiskit Nature

SaashaJoshi:
Resolves #154

PR #269, opened 2023-06-19
Move configuration out of `extremal-python-dependencies`

garrison:
This modifies the `extremal-python-dependencies` script so that it is configured on the command line.  This is one step toward making it a standard script.  Configuration is now provided by command line flags in the CI workflows.

### DIFF
--- a/.github/workflows/test_development_versions.yml
+++ b/.github/workflows/test_development_versions.yml
@@ -31,7 +31,11 @@ jobs:
         run: |
           python -m pip install --upgrade pip tox
           python -m pip install toml typer
-          python tools/extremal-python-dependencies.py pin-dependencies dev --inplace
+          python tools/extremal-python-dependencies.py pin-dependencies \
+              "qiskit-terra @ git+https://github.com/Qiskit/qiskit-terra.git" \
+              "qiskit-nature @ git+https://github.com/qiskit-community/qiskit-nature.git" \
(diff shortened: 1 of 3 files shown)


garrison:
Waiting on #230

caleb-johnson:
> This modifies the `extremal-python-dependencies` script so that it is configured on the command line. This is one step toward making it a standard script. Configuration is now provided by command line flags in the CI workflows.

Nice, is the idea that we include this in `tools` as something we can run outside the context of CI?

garrison:
> Nice, is the idea that we include this in `tools` as something we can run outside the context of CI?

That is a nice benefit of this, but actually, my immediate goal is to make this script more generic, so that we aren't configuring it _within_ itself.  This way, we can standardize this script among all the repositories that use it, or, better yet, spin it out into its own repository with tests and brief documentation.  From there, we could `pip install` it in the workflows that need it and maintain it only in one place.

Currently, forks of this code exist at
- https://github.com/qiskit-community/prototype-zne/blob/main/tools/extremal_dependency_versions.py
- https://github.com/Qiskit-Extensions/staged-primitives/blob/main/tools/extremal_dependency_versions.py
- and a few other places internally, not to mention packages that did it the old, `setup.py` way (quantum-prototype-template, qrao, and quantum kernel training)

caleb-johnson (approved):
LGTM thanks 👍

PR #274, opened 2023-06-19
Ignore opflow warnings

caleb-johnson:
This PR ignores *all warnings* from opflow, which is being prohibitively noisy in our unit tests. It brings our total warnings down from 104 to 14. This would ignore non-deprecation-warnings as well, but since opflow is deprecated and we do not use it directly in our source code, this is an acceptable tradeoff.

This PR also specifies the test path in the new pytest config in `pyproject.toml`, so it no longer needs to be specified in `tox.ini`, so we remove it in tox.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -111,3 +111,4 @@ omit = [
 [tool.pytest.ini_options]
 testpaths = ["./test/"]
 filterwarnings = ["ignore:::.*opflow*"]
+addopts = "--doctest-modules -rs --durations=10"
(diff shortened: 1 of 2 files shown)


garrison:
> This PR also specifies the test path in the new pytest config in `pyproject.toml`

There is also pytest config at the bottom of `tox.ini`.  We should choose one place and merge them together.

garrison:
Aside from my one comment, I think we should merge this.  It will be liberating to greatly reduce the warnings.  If we find an even _better_ way to do this, we can always back out this change at that time.

caleb-johnson:
> Aside from my one comment, I think we should merge this. It will be liberating to greatly reduce the warnings. If we find an even _better_ way to do this, we can always back out this change at that time.

OK I moved everything into pyproject since it has all the other tool configs

PR #276, opened 2023-06-21
Fix bug in decompose-qpd-instructions

caleb-johnson:
If `TwoQubitQPDGate` indices are specified out of order in the arguments to `decompose_qpd_instructions`, the  [data_id_offset](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/665196fc9b14f31d620e489e18fb339b33deb895/circuit_knitting/cutting/qpd/qpd.py#L429) variable specified in the function will not work as intended when decomposing those gates, and gates which are not `BaseQPDGate` instances will be treated as QPD gates, causing mysterious index errors.

This PR fixes that bug and implements a unit test to test against this problem.

### DIFF
--- a/releasenotes/notes/dx-qpd-inst-bug-fe9c50dad716b848.yaml
+++ b/releasenotes/notes/dx-qpd-inst-bug-fe9c50dad716b848.yaml
@@ -3,4 +3,4 @@ fixes:
   - |
     Fixed a bug in :func:`~circuit_knitting.cutting.qpd.decompose_qpd_instructions` which would
     cause index errors if :class:`~circuit_knitting.cutting.qpd.TwoQubitQPDGate` indices were
-    specified out of order.
+    not specified in ascending order.
--- a/releasenotes/notes/dx-qpd-inst-bug-fe9c50dad716b848.yaml
+++ b/releasenotes/notes/dx-qpd-inst-bug-fe9c50dad716b848.yaml
@@ -1,0 +1,6 @@
+---
+fixes:
+  - |
+    Fixed a bug in :func:`~circuit_knitting.cutting.qpd.decompose_qpd_instructions` which would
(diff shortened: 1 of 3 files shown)


test/cutting/qpd/test_qpd.py
```diff
@@ -170,7 +181,6 @@ def test_decompose_qpd_instructions(self):
                 e_info.value.args[0]
                 == "Each decomposition must contain either one or two elements. Found a decomposition with (0) elements."
             )
-
```
caleb-johnson, line 173: I keep fixing this in my PRs and then reverting it so you dont have to deal with it in code review haha...

I'm just going to leave this change in here and try to sneak it in ;)

releasenotes/notes/dx-qpd-inst-bug-fe9c50dad716b848.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    Fixed a bug in :func:`~circuit_knitting.cutting.qpd.decompose_qpd_instructions` which would
+    cause index errors if :class:`~circuit_knitting.cutting.qpd.TwoQubitQPDGate` indices were
+    specified out of order.
```
garrison, line 6:
```suggestion
    not specified in strictly ascending order.
```
This is more clear to me.  What do you think?

test/cutting/qpd/test_qpd.py
```diff
@@ -148,6 +148,17 @@ def test_decompose_qpd_instructions(self):
                 e_info.value.args[0]
                 == "The number of map IDs (1) must equal the number of decompositions in the circuit (2)."
             )
+        with self.subTest("test_unordered_indices"):
```
garrison, line 151: I don't think this really matters much, but I'll just note that that with the underscores, this doesn't really match the style of the other subtest names.

garrison (approved):
Thanks, this is great, and something nice to discover before our users.

PR #277, opened 2023-06-21
Add a supported_gates function

caleb-johnson:
There should be a central place where users can find out whether or not a gate is supported without using a `try` block and querying `qpdbasis_from_gate`. I've addressed this here with a simple function that returns a hard-coded `set`. I had also considered a function which takes a gate as input and returns a `bool`, but I just think this implementation is better for both the dev and user.

### DIFF
--- a/circuit_knitting/cutting/qpd/qpd.py
+++ b/circuit_knitting/cutting/qpd/qpd.py
@@ -215,14 +215,14 @@ def qpdbasis_from_gate(gate: Gate) -> QPDBasis:
         return f(gate)
 
 
-def supported_gates() -> tuple[str]:
+def supported_gates() -> set[str]:
     """
     Return a list of gate names which are supported for automatic decomposition.
 
     Returns:
         Tuple containing names of gates which are supported for automatic decomposition.
     """
-    return ("rxx", "ryy", "rzz", "crx", "cry", "crz", "cx", "cz")
+    return {"rxx", "ryy", "rzz", "crx", "cry", "crz", "cx", "cz"}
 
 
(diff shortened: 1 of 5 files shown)


circuit_knitting/cutting/qpd/qpd.py
```diff
@@ -215,6 +215,16 @@ def qpdbasis_from_gate(gate: Gate) -> QPDBasis:
         return f(gate)
 
 
+def supported_gates() -> set[str]:
```
garrison, line 218: I think we should also export this from `__init__.py` in the same directory.

circuit_knitting/cutting/qpd/qpd.py
```diff
@@ -215,6 +215,16 @@ def qpdbasis_from_gate(gate: Gate) -> QPDBasis:
...
+    """
+    Return a set of gate names supported for automatic decomposition.
+
+    Returns:
+        Set of gate names supported for automatic decomposition.
+    """
+    return {"rxx", "ryy", "rzz", "crx", "cry", "crz", "cx", "cz"}
```
garrison, line 225:
```suggestion
    return set(_qpdbasis_from_gate_funcs)
```

I tested this locally, and it works.  Doing this the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) way, as I suggest, will make sure this list does not go out of date.

caleb-johnson: Thank you! I've been racking my brain trying to figure out how to make our test fail if a gate is added and it's not updated

PR #283, opened 2023-06-22
Correct module import in docs/conf.py

SaashaJoshi:


garrison:
Thank you!  Previously, we were implicitly relying on `importlib_metadata` being installed as a transitive dependency.  This was causing an `ImportError` for @SaashaJoshi.  Since we only support Python 3.8 and higher now, we can just use the `importlib` that is built in to Python beginning with version 3.8.

garrison (approved):

PR #286, opened 2023-06-23
Add explicit numpy dependency

garrison:
We are currently relying on numpy being installed as a transitive dependency (through, at the very least, Qiskit Terra).  Even better would be to specific it explicitly, and to set the minimum version following https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/70#issuecomment-1501957148.

#### TODO
- [x] add a release note

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,7 @@ classifiers = [
 requires-python = ">=3.8"
 
 dependencies = [
+    "numpy>=1.23.0",
     "qiskit-aer>=0.12.0",
     "qiskit-terra>=0.24.0",
     "qiskit-nature>=0.6.0",
--- a/releasenotes/notes/explicit-numpy-dependency-fdd89846b37a8d4c.yaml
+++ b/releasenotes/notes/explicit-numpy-dependency-fdd89846b37a8d4c.yaml
@@ -1,0 +1,7 @@
+---
+upgrade:
+  - |
+    Numpy 1.23.0 or later is now required.  The :func:`~numpy.kron`
+    method in earlier versions `has known performance issues
+    <https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/70#issuecomment-1500514513>`__,
+    and this method is used heavily by the CutQC wire cutting module.


caleb-johnson:
Isn't leaving it unpinned a little more robust since we don't rely on any unstable numpy features?

garrison:
> Isn't leaving it unpinned a little more robust since we don't rely on any unstable numpy features?

What do you mean by "more robust"?

I'll say a few things, though.

1. We should specify `numpy` as dependency since we import it directly.  We should not rely on it being transitively depended on by Qiskit, even though there is no real risk of that changing.
2. Since we depend on it, it'd be nice to pin the minimum version to something sensible.  Qiskit Terra depends on `numpy>=1.17`, so we _could_ follow suit and do the same.
3. We know that the performance degrades on numpy versions earlier than 1.23 -- at least, if one uses CutQC to reconstruct the full probability distribution.  I considered adding a warning, but it seems unlikely that everyone would notice it.  I could think of no harm to bumping the minimum version of numpy to 1.23, so that's what I did here.

caleb-johnson:
> > Isn't leaving it unpinned a little more robust since we don't rely on any unstable numpy features?
> 
> What do you mean by "more robust"?
> 
> I'll say a few things, though.
> 
> 1. We should specify `numpy` as dependency since we import it directly.  We should not rely on it being transitively depended on by Qiskit, even though there is no real risk of that changing.
> 2. Since we depend on it, it'd be nice to pin the minimum version to something sensible.  Qiskit Terra depends on `numpy>=1.17`, so we _could_ follow suit and do the same.
> 3. We know that the performance degrades on numpy versions earlier than 1.23 -- at least, if one uses CutQC to reconstruct the full probability distribution.  I considered adding a warning, but it seems unlikely that everyone would notice it.  I could think of no harm to bumping the minimum version of numpy to 1.23, so that's what I did here.

Thanks, I like the justification of "we import it directly, so it is *our* dependency".

I agree, let's pin minimum to 1.23

caleb-johnson (approved):
LGTM 👍 thanks!

PR #287, opened 2023-06-24
Cancel GH Action workflow when a PR is updated

garrison:
This change is intended to make it so GitHub Actions will cancel a workflow that belongs to a PR if that PR has since already been updated.  In this case, there's typically no benefit to having the run finish.  In the case of something triggering these workflows that is _not_ a PR, the concurrency group will be given as the commit digest, so it will run fully for every commit.  This way, CI will run on every push to `main` and the stable branches.


garrison:
Ugh, the annoying part of this is that I got an email for every workflow with cancelled jobs (6 of them, in total)

PR #289, opened 2023-06-26
Add explicit dependency on rustworkx and scipy

garrison:
This is similar to #286 in that it adds an explicit dependency on something that we are already importing directly (in this case, rustworkx).  I set the minimum version to 0.12.0 since that is what Qiskit Terra 0.24.0 requires.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,6 +27,7 @@ requires-python = ">=3.8"
 
 dependencies = [
     "numpy>=1.23.0",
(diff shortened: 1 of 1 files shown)


garrison:
We can removed the upper bound on scipy once https://github.com/pyscf/pyscf/pull/1771 makes it into a pyscf release.

caleb-johnson (approved):
Awesome, thanks. Hopefully pyscf will update soon 🤞

PR #296, opened 2023-06-30
`CSXGate`

garrison:
I started with #295, performed a small refactor, and moved the T gate from the second qubit to the first.


caleb-johnson (approved):
Nice, I'm still not sure why this wasn't working in my branch when I tried prepending TGates to qubit 0...

PR #298, opened 2023-06-30
Support CPhaseGate, CSGate and CSdgGate

caleb-johnson:
Support automatic, optimal decomposition of CPhaseGate, CSGate and CSdgGate.

Add a few minor things in release note and qpd docstring for `CSXGate` support I missed when merging #296  (sorry @garrison  :) )


circuit_knitting/cutting/qpd/qpd.py
```diff
@@ -315,11 +318,24 @@ def _(gate: RXXGate | RYYGate | RZZGate | CRXGate | CRYGate | CRZGate):
...
+    if gate.name in {"csx", "cs"}:
+        for operations in unique_by_id(m[0] for m in retval.maps):
+            operations.insert(0, TGate())
+    elif gate.name in {"csdg"}:
+        for operations in unique_by_id(m[0] for m in retval.maps):
+            operations.insert(0, TdgGate())
+    else:
+        assert gate.name in {"cp"}
+        for operations in unique_by_id(m[0] for m in retval.maps):
+            operations.insert(0, PhaseGate(theta / 2))
```
garrison, line 337: This would have fewer branches and I actually think this would be easier to reason about if it were at least two distinct methods.  `cs`, `csdg`, and `cp` might naturally all go together since they are controlled rotations about the Z axis, but I think if will be more clear if `csx` remains its own method, at the very least.

garrison (commented):
Thanks, it will be exciting to support these. :tada:

circuit_knitting/cutting/qpd/qpd.py
caleb-johnson: > This would have fewer branches and I actually think this would be easier to reason about if it were at least two distinct methods. `cs`, `csdg`, and `cp` might naturally all go together since they are controlled rotations about the Z axis, but I think if will be more clear if `csx` remains its own method, at the very least.

Agree, I think breaking this out into 3 functions looks pretty natural.

garrison (approved):
Thanks, this looks great. :tada:

PR #294, opened 2023-06-30
Add support for `SwapGate`, `iSwapGate`, and `DCXGate`

garrison:
I might clean this up a bit, but hey -- it works. :slightly_smiling_face: 

#### Remaining action items

- [x] fix coverage (one line is currently missing)
- [x] clean up & add comments
- [x] add to release note (might need to rebase after #292 is merged to please reno)
- [x] possibly add additional tests of `kappa` values from other recent gates
- [x] update supported gates in docstring

Issue #258, opened 2023-06-16
Decomposition of gates with 3 or more qubits


garrison:
We may eventually want to perform gate cutting on gates that act on 3+ qubits: for instance, `CSwapGate`, `CCXGate`, `C3XGate`, etc.  In each case, we could start by _only_ support cutting such a gate into two pieces, as that is complex enough to handle and will cover the main efficiency gains that could come from doing this.

- The three-qubit gates could be split one of two ways: either ABB or AAB.
- The `CNXGates` can be split in `N - 1` different ways, depending on how many control bits end up split on the same side as the target bit.

Of course, in addition to refactoring `QPDBasis`, we'll need to derive what the coefficients of the decomposition are in each of the above cases we want to support.

PR #302, opened 2023-07-02
Implement cutting of general 2-qubit unitaries

### LINKED ISSUE #186: Implement LO cutting of general two-qubit unitaries
The method is presented in https://arxiv.org/abs/2006.11174v2.

The class `TwoQubitWeylDecomposition` at https://github.com/Qiskit/qiskit-terra/blob/main/qiskit/quantum_info/synthesis/two_qubit_decompose.py appears to do the decomposition we need, given as Eq. (6) in this paper.

garrison:
Builds on #294.  Closes #186.  ~~I'm marking as "on hold" at least until we merge the other explicit gate PRs.~~

#### Remaining action items

- [x] Release note
- [x] Rename `supported_gates()` -> `explicitly_supported_gates()`
- [x] update docstring
- [x] More comments/explanation
- [x] Update references in readme and docs


circuit_knitting/cutting/qpd/qpd.py
```diff
@@ -225,10 +227,25 @@ def qpdbasis_from_gate(gate: Gate) -> QPDBasis:
     try:
         f = _qpdbasis_from_gate_funcs[gate.name]
     except KeyError:
-        raise ValueError(f"Gate not supported: {gate.name}") from None
+        pass
     else:
         return f(gate)
 
+    if isinstance(gate, Gate) and gate.num_qubits == 2:
```
caleb-johnson, line 234: YES!!

caleb-johnson (approved):
So happy you figured this all out. I'm approving bc tests look good, but I'll be interested to hear how you connected the dots between the literature and the WeylDecomposer class in Qiskit. Thanks!

caleb-johnson:
Maybe not this PR, but should we provide a lookup table for Qiskit gates --> Sampling overhead?

garrison:
> Maybe not this PR, but should we provide a lookup table for Qiskit gates --> Sampling overhead

I agree -- it is worth having a section about this in the documentation.  I could imagine this going in either the "explanation" section or the API references.  It feels like a "reference"-type documentation to me, even though it is not strictly about the API.

>Also, we now have a grab bag of optimal and non-optimal decomps in our supported gate set. How do we make this clear? Maybe two sections in the lookup table?

The authors of https://arxiv.org/abs/2006.11174v2 left open the possibility that their decomposition might not actually be optimal.  (They discuss this in Sec. 2.4.)  However, corollary 4.4 of https://arxiv.org/abs/2205.00016v2 actually shows that each of the gates that we explicitly support has an optimal decomposition.  So, we are really left with two classes: decompositions that we can _prove_ are optimal, and decompositions that _may_ be optimal.  Pretty much everything we explicitly mention in our table is going to fit in the first category; I can't think of a single "standard" gate that actually fits in the second.

garrison:
We might as well put this on the 0.3 milestone, right?  Deciding which release will influence whether I modify existing release notes or add new ones.

caleb-johnson:
> We might as well put this on the 0.3 milestone, right? Deciding which release will influence whether I modify existing release notes or add new ones.

Yes I thnk so

circuit_knitting/cutting/__init__.py
```diff
@@ -57,7 +57,7 @@
     qpd.generate_qpd_weights
     qpd.generate_qpd_samples
     qpd.decompose_qpd_instructions
-    qpd.supported_gates
+    qpd.explicitly_supported_gates
```
caleb-johnson, line 60: I wanted to suggest `optimally_supported_gates`, but I can't convince myself I like it more. I think they are all optimal, at least?

garrison: Yeah, they are all optimal -- even the ones for which we do not have explicit support.

Another option could be to get rid of this function -- or at least to make it private.  The reason it was introduced in #277 was to support #278, but now that _all_ two-qubit gates can be cut, it's no longer necessary.  It might be nice to have something like this if we one day support some (but not all) 3-qubit gates (#258), which could be a case for keeping it but prefixing it with an underscore.  This way, we'd at least have a use case in mind (again) before committing to support it.

By the way, all _essential_ tests (i.e., all but a few specific ones) work if we take this PR and remove all the explicit gate support, i.e., if we apply the further patch:

```diff
diff --git a/circuit_knitting/cutting/qpd/qpd.py b/circuit_knitting/cutting/qpd/qpd.py
index 1e5be49..c6974a8 100644
--- a/circuit_knitting/cutting/qpd/qpd.py
+++ b/circuit_knitting/cutting/qpd/qpd.py
@@ -573,13 +573,6 @@ def qpdbasis_from_gate(gate: Gate) -> QPDBasis:
         ValueError: Cannot decompose gate with unbound parameters.
         ValueError: ``to_matrix`` conversion of two-qubit gate failed.
     """
-    try:
-        f = _qpdbasis_from_gate_funcs[gate.name]
-    except KeyError:
-        pass
-    else:
-        return f(gate)
-
     if isinstance(gate, Gate) and gate.num_qubits == 2:
         try:
             mat = gate.to_matrix()
```

so explicit support doesn't mean a ton anyway.

garrison: The more I thought about it, the more I felt like the most reasonable thing to do is to remove this function from the public API for now, hence my change in 8e95c4acdc369b8535682521da77d6cb810822b6.

circuit_knitting/cutting/qpd/__init__.py
```diff
@@ -18,7 +18,6 @@
     decompose_qpd_instructions,
     WeightType,
     qpdbasis_from_gate,
-    supported_gates,
```
caleb-johnson, line 21: Is there a 2-qubit gate in Qiskit that would break our cutting workflow if we tried to cut it?

I'm wondering if we can replace
`if g in supported_gates():`

with
`if g.num_qubits == 2:`

caleb-johnson: I now see your docstring update below. They have to implement `to_matrix`

circuit_knitting/cutting/qpd/qpd.py
```diff
@@ -619,6 +630,54 @@ def _copy_unique_sublists(lsts: tuple[list, ...], /) -> tuple[list, ...]:
     return tuple(copy_by_id[id(lst)] for lst in lsts)
 
 
+def _u_from_thetavec(
+    theta: np.typing.NDArray[np.float64] | Sequence[float], /
+) -> np.typing.NDArray[np.complex128]:
+    r"""
+    Exponentiate the non-local portion of a KAK decomposition.
```
caleb-johnson, line 637: Great writeup here, thanks

caleb-johnson (approved):
Nice work and great documentation, thanks! LGTM 👍

PR #303, opened 2023-07-03
Add support for `SingleQubitQPDGate` to be passed through the transpiler

garrison:
This is a first step toward making our workflow more flexible with regard to transpilation.  I created a new file for tests like this one.  The `_directive` property is necessary for this test to pass.

### DIFF
--- a/circuit_knitting/cutting/qpd/instructions/qpd_gate.py
+++ b/circuit_knitting/cutting/qpd/instructions/qpd_gate.py
@@ -201,6 +201,11 @@ class SingleQubitQPDGate(BaseQPDGate):
             qc.append(CircuitInstruction(op, [qc.qubits[0]], []))
         self.definition = qc
 
+    @property
+    def _directive(self):
+        """``True`` if the ``basis_id`` is unassigned, which implies this instruction cannot be decomposed."""
+        return self.basis_id is None
+
     def __eq__(self, other):
         """Check equivalence for SingleQubitQPDGate class."""
         return super().__eq__(other) and self.qubit_id == other.qubit_id
--- a/test/cutting/test_cutting_workflows.py
+++ b/test/cutting/test_cutting_workflows.py
@@ -1,0 +1,50 @@
+# This code is a Qiskit project.
+
+# (C) Copyright IBM 2023.
+
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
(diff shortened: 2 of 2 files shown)


circuit_knitting/cutting/qpd/instructions/qpd_gate.py
```diff
@@ -201,6 +201,11 @@ def _define(self) -> None:
             qc.append(CircuitInstruction(op, [qc.qubits[0]], []))
         self.definition = qc
 
+    @property
```
caleb-johnson, line 204: Could you please just quickly explain this mechanism here? Not familiar with _directive

garrison: It is something that @mtreinish told me about.  It tells the transpiler to treat it as something special that ought to be passed through most passes without being touched, and is used for `Barrier`s and a few other things.

See [this brief description](https://github.com/Qiskit/qiskit-terra/blob/e9f8b7c50968501e019d0cb426676ac606eb5a10/qiskit/circuit/instruction.py#L55-L57), which itself links to https://github.com/Qiskit/qiskit-terra/issues/5811

test/cutting/test_cutting_workflows.py
```diff
@@ -0,0 +1,50 @@
...
+    # Create a fake backend, and modify the target gate set so it thinks a
+    # SingleQubitQPDGate is allowed.
+    backend = FakeLagosV2()
+    backend = deepcopy(backend)
+    sample_qpd_instruction = SingleQubitQPDGate(QPDBasis.from_gate(CXGate()), 1)
+    backend.target.add_instruction(
+        sample_qpd_instruction,
+        {(i,): None for i in range(backend.target.num_qubits)},
+    )
+    pass_manager = generate_preset_pass_manager(3, backend)
```
mtreinish, line 44: You can do something like:

```suggestion
    target = deepcopy(backend.target)
    sample_qpd_instruction = SingleQubitQPDGate(QPDBasis.from_gate(CXGate()), 1)
    target.add_instruction(
        sample_qpd_instruction,
        {(i,): None for i in range(target.num_qubits)},
    )
    pass_manager = generate_preset_pass_manager(3, target=target)
```

backend's are not guaranteed to be serializable with pickle (which is what deepcopy does by default) but a target is.

garrison: Thank you, @mtreinish!

caleb-johnson (approved):
Good thinking getting this started, thanks 👍

PR #304, opened 2023-07-03
Return more info about circuit separation during partitioning

caleb-johnson:
This is the first step toward supporting an easier interface for updating parameters associated with cut gates.

In the current cutting workflow, users are given a dictionary of subcircuits and the bases associated with each unique cut. Instead of only returning the bases, we should return which gates in which subcircuits each of the bases are associated with.

Once this information is available from `partition_problem`, implementing functions for manipulating the QPD gates in circuits will be more straightforward.

This is currently implemented as a breaking change, as the `PartitionedCuttingProblem.bases` field is now the `cuts` field. Users should access the bases via the new field. It is probably preferable to deprecate the `bases` field


circuit_knitting/cutting/cutting_decomposition.py
```diff
@@ -221,14 +229,28 @@ def partition_problem(
     bases = []
     i = 0
     for inst in qpd_circuit.data:
+        if isinstance(inst.operation, SingleQubitQPDGate):
```
caleb-johnson, line 232: We probably want to support these for wire cutting, but they are not well-handled as of this PR, so this ValueError is appropriate. We can remove it for the wire cutting PR, if necessary

test/cutting/test_backwards_compatibility.py
```diff
@@ -51,9 +51,10 @@ def test_v0_2_cutting_width_workflow():
     circuit = EfficientSU2(4, entanglement="linear", reps=2).decompose()
     circuit.assign_parameters([0.8] * len(circuit.parameters), inplace=True)
     observables = PauliList(["ZZII", "IZZI", "IIZZ", "XIXI", "ZIZZ", "IXIX"])
-    subcircuits, bases, subobservables = partition_problem(
+    subcircuits, cuts, subobservables = partition_problem(
         circuit=circuit, partition_labels="AABB", observables=observables
     )
+    bases = [cut_info.basis for cut_info in cuts]
```
caleb-johnson, line 57: This PR is a breaking change, users now need to get each basis from the `CutInfo` object

circuit_knitting/cutting/cutting_decomposition.py
```diff
@@ -28,17 +28,24 @@
...
-    bases: list[QPDBasis]
+    cuts: list[CutInfo]
     subobservables: dict[str | int, QuantumCircuit] | None = None
 
 
+class CutInfo(NamedTuple):
+    """The decomposition and location information associated with one cut."""
+
+    basis: QPDBasis
+    gates: list[tuple[Hashable, int]] | list[int] | int
```
caleb-johnson, line 46: This data tuple can be used to describe multiple gates spread across several subcircuits, multiple gates within a single circuit, and one gate within a single circuit

caleb-johnson:
@garrison any comment on breaking change vs deprecating the old field?

circuit_knitting/cutting/cutting_reconstruction.py
```diff
@@ -72,7 +72,7 @@ def reconstruct_expectation_values(
         for label, subobservable in observables.items():
             if any(obs.phase != 0 for obs in subobservable):
                 raise ValueError("An input observable has a phase not equal to 1.")
-        subobservables_by_subsystem = observables
+        subobservables_by_subsystem = observables # type: ignore
```
caleb-johnson, line 75: I've moved everything in `decomposition` module to Hashable. This line is throwing a mypy error because the expression has keys `str | int`, and the variable is expecting `Hashable`. This seems like a false positive to me, so I've commented it out. This is "more correct" than the line we had commented out in the other module, but I still kind of wish we had just done `str | int` all the way through the `qpd` module. I don't really believe supporting `Hashable` is all that great...being able to sort your keys is better.

circuit_knitting/cutting/cutting_reconstruction.py
```diff
@@ -72,7 +72,7 @@ def reconstruct_expectation_values(
         for label, subobservable in observables.items():
             if any(obs.phase != 0 for obs in subobservable):
                 raise ValueError("An input observable has a phase not equal to 1.")
-        subobservables_by_subsystem = observables
+        subobservables_by_subsystem = observables  # type: ignore
```
caleb-johnson, line 75: I believe this is a false positive from mypy. The variable is a `dict[Hashable, Any]`, and the expression is a `dict[str | int, Any]`. I believe the expression lies in a subset of the variable, which should be fine.

garrison: https://mypy.readthedocs.io/en/stable/generics.html#variance-of-generic-types explains why `dict` is considered an invariant type rather than a covariant type.  But yes, I think ignoring this makes the most sense for us for now.

caleb-johnson: Cool, since this is technically correct, this means we have "correct" type hints with respect to `Hashable` and `str | int`. All the functions correctly indicate what functionality they expect

Issue #215, opened 2023-05-30
Configure Intersphinx to link to classes of dependencies' API docs


garrison:
See https://docs.readthedocs.io/en/stable/guides/intersphinx.html

It seems that intersphinx is already used by some other Qiskit projects, including qiskit-experiments, qiskit-machine-learning, qiskit-nature, qiskit-optimization, qiskit-terra, and retworkx.

caleb-johnson:
@garrison did #308 resolve this?

garrison:
> @garrison did #308 resolve this?

It is the first step.  As a next step, @SaashaJoshi is doing an audit to make sure all the external class links actually work, rather than the few big ones that that PR got working (QuantumCircuit and PauliList).  (I _think_ every class has to be marked as `:external` at least once for it to be picked up, but Saasha will probably have a better understanding.)

PR #308, opened 2023-07-05
Configure Intersphinx to docs

SaashaJoshi:
Related to #215


garrison (approved):
Thanks!  I think we should just merge this immediately, and you can work on auditing the existing links in a subsequent PR.  This is already a significant improvement.

PR #311, opened 2023-07-06
Add documentation links to qiskit-nature and qiskit-ibm-runtime for intersphinx

SaashaJoshi:


garrison:
Does doing this alone allow any links to work that did not work before?  I am curious.  (We might have to prefix each identifier once with `:external` for it to be picked up.)

SaashaJoshi:
> Does doing this alone allow any links to work that did not work before? I am curious. (We might have to prefix each identifier once with `:external` for it to be picked up.)

I realized we do not need `:external` anywhere. The interlinking works without it as well (I will make this change in the next commit). 
However, I see that the internal classes like `:class:`TwoQubitQPDGate`` do not get linked; I am still figuring out this one.

garrison:
> However, I see that the internal classes like `:class: TwoQubitQPDGate` do not get linked; I am still figuring out this one.

It _might_ be enough to say ``` :class:`.TwoQubitQPDGate` ```, i.e., to prepend it with a dot each time it is mentioned.

circuit_knitting/forging/entanglement_forging_operator.py
```diff
@@ -22,9 +22,9 @@
 class EntanglementForgingOperator:  # noqa: D301
     r"""Operator class for Entanglement Forging.
 
-    A class that contains the :math:`2N` qubit Pauli operator :math:`\hat{O} = \sum_{i, j} w_{i, j} \hat{T}_{i, j} \otimes \sum_{a, b} \hat{S}_{a, b}`
-    and associated weights. These operators are knitted by the :class:`EntanglementForgingKnitter` to provide esimates of the
-    energy for the :class:`EntanglementForgingVQE`.
+    A class that contains the :math:`2N` qubit Pauli operator :math:`\hat{O} = \sum_{i, j} w_{i, j} \hat{T}_{i,
+    j} \otimes \sum_{a, b} \hat{S}_{a, b}` and associated weights. These operators are knitted by the
+    :class:`.EntanglementForgingKnitter` to provide estimates of the energy for the :class:`.EntanglementForgingVQE`.
```
garrison, line 27: > * Reference to non-existing `EntanglementForgingVQE` class in `entanglement_forging_operator` (line 27).

`grep` to the rescue: I realized the same identifier mistakenly appears at 
https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/34b9fea053b1784c920c2401fe0a32dbc7a94b7a/test/forging/test_entanglement_forging_ground_state_solver.py#L12

So it looks like we should reference `EntanglementForgingGroundStateSolver` above, and might as well update that line, too.

SaashaJoshi: Awesome. This makes sense.

circuit_knitting/utils/observable_grouping.py
```diff
@@ -125,22 +125,20 @@ def most_general_observable(
...
+    r"""Set of mutually qubit-wise commuting observables."""
 
+    #: A single Pauli string that contains all qubit-wise measurements needed
+    #: to measure everything in ``commuting_observables``.
     general_observable: Pauli
+
+    #: Observables that can be measured simultaneously
     commuting_observables: list[Pauli]
+
+    #: The indices of non-identity :class:`~qiskit.quantum_info.Pauli`\ s in ``general_observable``
```
garrison, line 137:
```suggestion
    #: The indices of non-identity :class:`~qiskit.quantum_info.Pauli`\ s in ``general_observable``.
```

garrison:
> Related to:
> #215

I think this may actually _close_ #215.  Is there anything left to do after this?

SaashaJoshi:
I also made edit for #216 . I realized that issue was also covered in this PR.

PR #313, opened 2023-07-10
Switch to qiskit-sphinx-theme

### LINKED ISSUE #312: Upgrade sphinx theme
[qiskit-sphinx-theme](https://github.com/Qiskit/qiskit_sphinx_theme) v1.13 introduces a new theme built on top of [furo](https://pypi.org/project/furo/). Upgrade the ckt sphinx theme to the new qiskit theme.

[furo theme preview](https://qiskit-sphinx-theme-pr-441.134fzbwmnxci.us-south.codeengine.appdomain.cloud/)

Eric-Arellano:
Resolves #312 

Switches to https://github.com/Qiskit/qiskit_sphinx_theme. The theme is based on Furo, which is used for pip docs, and is a much more functional and modern theme than the current Alabaster.

Note that qiskit-sphinx-theme will add a dedicated Ecosystem theme in late July/August. The current theme is the same as Qiskit docs. This is still some forward progress, though, so no need to wait.


tox.ini
```diff
@@ -60,4 +60,4 @@ extras =
 commands =
   python -c 'import shutil, pathlib; shutil.rmtree(pathlib.Path("docs") / "stubs", ignore_errors=True)'
   python -c 'import shutil, pathlib; shutil.rmtree(pathlib.Path("docs") / "_build" / "html" / ".doctrees", ignore_errors=True)'
-  sphinx-build -j auto -b html -W -T --keep-going {posargs} docs/ docs/_build/html
+  sphinx-build -j auto -W -T --keep-going {posargs} docs/ docs/_build/html
```
garrison, line 63: I just looked this up.  It looks like `-b html` [is the default](https://www.sphinx-doc.org/en/master/man/sphinx-build.html).  Is there a reason it shouldn't be set explicitly?

Eric-Arellano: Exactly. The code will work just the same, so either approach is fine.

As a general philosophy, I bias much more towards not setting options that are the default. It's added noise. Defaults exist for a reason. When you rely on defaults, it makes it much clearer where you're diverging.

But either is fine and I can revert this if you have a preference.

garrison: > It's added noise.

Modifying a line that is not necessary, with no explanation, is also added noise.  But the change is fine -- now that we've discussed it, let's keep it.

garrison (approved):
Thanks!  I looked through the build, and it looks great. :tada:

tox.ini
Eric-Arellano: For sure. For context, I was going to add an explanation once the PR was marked ready for review (it had `[wip]`). I review my own PRs before asking for reviews

PR #314, opened 2023-07-10
Upload docs build in CI

Eric-Arellano:
This is useful to iterate on docs and debug CI issues.

The artifact will be uploaded to GitHub Actions like this:

Screenshot 2023-07-10 at 7 56 47 AM

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -43,7 +43,6 @@ jobs:
         with:
           name: html_docs
           path: ./docs/_build/html
-
       - name: Deploy docs
         if: ${{ github.ref == 'refs/heads/stable/0.2' }}
         uses: peaceiris/actions-gh-pages@v3
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -37,6 +37,13 @@ jobs:
         shell: bash
         run: |
           tox -edocs
+      - name: Upload docs artifact
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: html_docs
+          path: ./docs/_build/html
+
       - name: Deploy docs
         if: ${{ github.ref == 'refs/heads/stable/0.2' }}
         uses: peaceiris/actions-gh-pages@v3


.github/workflows/docs.yml
```diff
@@ -37,6 +37,13 @@ jobs:
         shell: bash
         run: |
           tox -edocs
+      - name: Upload docs artifact
+        if: always()
```
caleb-johnson, line 41: Any reason you like the if statement here?

.github/workflows/docs.yml
```diff
@@ -37,6 +37,13 @@ jobs:
...
+        uses: actions/upload-artifact@v3
```
caleb-johnson, line 42: Does this allow for the docs to be built in parallel, or does it serve some other purpose?

caleb-johnson: *sigh* I should've read PR description. Disregard :)

Eric-Arellano: No. All this does is that after the docs are built, we upload them to GitHub CI so that users can download the docs and see how they look on their local machine.

You already have parallel docs enabled with the `-j auto`:

https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/c5207b191b601223e8e4e5eb56b2200c256101e7/tox.ini#L63

.github/workflows/docs.yml
caleb-johnson: I think we can just remove the if statement here

caleb-johnson (commented):
LGTM thanks 👍 

I'll let @garrison take a look before approving

.github/workflows/docs.yml
```diff
@@ -37,6 +37,13 @@ jobs:
...
+        with:
+          name: html_docs
+          path: ./docs/_build/html
+
```
garrison, line 46:
```suggestion
```

Eric-Arellano:
>  Is this artifact only useful for PRs, or is it useful for all docs builds?

Yeah, it's useful for all CI builds. And there is little downside to this PR other than the couple lines of extra code. The upload should be fast and your docs are only 25MB.

Concrete example: we make a big styling change in https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/313. I want to make it easy for us all to see what the new docs will look like, without making you build the docs locally.

garrison (approved):

.github/workflows/docs.yml
caleb-johnson: I see from @garrison comment below, the if: always is in place to force a broken build to upload

PR #307, opened 2023-07-05
Support freezing Hartree-Fock energies at each forging iteration

caleb-johnson:
This PR allows users to set the HF energy explicitly, rather than calculating it at each iteration with quantum experiments. This can improve the runtime of the algorithm, and it can also improve results.

- [x] Make release note
- [x] Link to explanatory in docstring
- [x] Change variable names in knitter to reflect that the hf_energy has been shifted
- [x] Get new code up to 100% coverage

PR #317, opened 2023-07-11
Run slow tests for coverage only

caleb-johnson:
#307 includes a slow test that also provides important code coverage. I suggest we run slow tests on coverage only. This will allow us to continue to test a higher proportion of our code without bogging down every workflow with slow tests.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -47,8 +47,7 @@ extras =
   test
   cplex
 commands =
-  coverage3 run --source circuit_knitting --parallel-mode -m pytest test/ --coverage {posargs}
-  coverage3 run --source circuit_knitting --parallel-mode -m pytest -m slow test/ --coverage {posargs}
+  coverage3 run --source circuit_knitting --parallel-mode -m pytest --run-slow test/ --coverage {posargs}
   coverage3 combine
   coverage3 html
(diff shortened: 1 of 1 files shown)


garrison:
> I suggest we run slow tests on coverage only.

To clarify: this change makes it so coverage run the slow tests by default when invoked by _tox_; the other tox test environments don't run them.  But within CI, the workflows explicitly opt for the other environments to run slow tests as well (see `--run-slow` in the `test_*_versions.yml` files).  I think this is fine, just noting it.

garrison (approved):

garrison:
Coverage takes a long time to run now, as there is a single test (test_asymmetric_bitstrings_CN) that takes 39 seconds on my machine.  I think it might be better to annotate the specific slow functions we actually _want_ to count coverage for, and run only those slow tests during coverage.  It might be annoying to figure out the details of this, but I'll probably do it one day when I am feeling especially annoyed by the status quo :slightly_smiling_face:

caleb-johnson:
> Coverage takes a long time to run now, as there is a single test (test_asymmetric_bitstrings_CN) that takes 39 seconds on my machine. I think it might be better to annotate the specific slow functions we actually _want_ to count coverage for, and run only those slow tests during coverage. It might be annoying to figure out the details of this, but I'll probably do it one day when I am feeling especially annoyed by the status quo 🙂

OK this had crossed my mind, but now that I see how bad the CN test is again, I'll look into it. Thanks

PR #318, opened 2023-07-12
Add caption headers to site table of contents

Eric-Arellano:
This keeps the same order of docs, but adds section headers so it's more obvious what is what.

It also adds a page to `<self>` so that you can go back to the home page.

Before:

Screenshot 2023-07-12 at 10 24 47 AM

After:

Screenshot 2023-07-14 at 9 54 02 AM

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -35,8 +35,6 @@ Contents
 
   About Circuit Knitting Toolbox <self>
   Installation Instructions <install>
-  API References <apidocs/index>
-  Release Notes <release-notes>
 
 .. toctree::
   :maxdepth: 2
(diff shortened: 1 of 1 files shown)


garrison:
I'm open to this, but it's bit unfortunate that the API references would then come before the tutorials, explanation, and how-to guides.

caleb-johnson:
> I'm open to this, but it's bit unfortunate that the API references would then come before the tutorials, explanation, and how-to guides.

I'm normally going straight to API when I go to docs, so maybe not *strictly* unfortunate. Also, the API contains info on all the tools, so I don't think it's so bad up there with the CKT blurb and Installation

Eric-Arellano:
I'm happy with whatever order you all think makes sense! And feel free to push commits directly to this branch.

caleb-johnson (dismissed):

Eric-Arellano:
@garrison, one option is to have a 3rd category group called `API` and then it has the entry underneath called `API References`. I like that because `API` is huge when expanded. So when expanded, you avoid covering the rest of the ToC entries. Wdyt?

garrison (approved):
This looks great, thanks.

PR #320, opened 2023-07-12
Use getters/setters internally in forging

caleb-johnson:
We have generally used the private class fields within class methods, rather than using the getters, but this is causing unnecessary dropped coverage in EF. This PR switches the EF modules to use getters/setters internally, in order to get credit for that code coverage.

### DIFF
--- a/circuit_knitting/forging/entanglement_forging_ansatz.py
+++ b/circuit_knitting/forging/entanglement_forging_ansatz.py
@@ -79,9 +79,9 @@ class EntanglementForgingAnsatz:
                     "There must be the same number of V bitstrings as U bitstrings."
                 )
 
-        self._circuit_u: QuantumCircuit = circuit_u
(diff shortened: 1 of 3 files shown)


garrison:
> Actually this PR somehow decreases the coverage.

Are you sure?   I think it is mainly complaining that _some_ of the lines you modified aren't covered (which is no surprise, since they weren't before, etiher).  If I compare the number of covered lines here to the base build, this one is higher.

caleb-johnson:
> > Actually this PR somehow decreases the coverage.
> 
> Are you sure? I think it is mainly complaining that _some_ of the lines you modified aren't covered (which is no surprise, since they weren't before, etiher). If I compare the number of covered lines here to the base build, this one is higher.

I clicked the coveralls links and it looked like 91.2 vs 91.7 before...maybe I misread it? I was a bit mystified by the decrease tbh...I'd re-open this for that .5% lol

PR #325, opened 2023-07-14
Cplex is now compatible with Python 3.11

garrison:
There are new cplex binaries for Python 3.11, so we can run all our tests on Python 3.11 now.

Ref https://github.com/qiskit-community/qiskit-optimization/pull/533


caleb-johnson:
Could we update the last section in INSTALL.md to reflect CPLEX now supports py311?

garrison:
> Could we update the last section in INSTALL.md to reflect CPLEX now supports py311?

Thanks -- yeah, it's good to have all these changes in the same PR, because we'll be looking back at this one again when Python 3.12 comes out in October, and cplex doesn't support _it_ yet, then. :slightly_smiling_face:

PR #326, opened 2023-07-14
Add CutWire instruction class

SaashaJoshi:
- [x] Add `CutWire` instruction class that denotes the location of a wire cut in a circuit.
- [x] Add `transform_to_move` function that transforms the `cut_wire` instruction to a `move` instruction.

Proposed functionality of the above-mentioned function:
  Input: QuantumCircuit w/ n qubits and m CutWire instructions
  Output: new QuantumCircuit, w n+m qubits, where every CutWire instruction has been replaced by a Move instruction on two qubits


circuit_knitting/cutting/qpd/instructions/cut_wire.py
```diff
@@ -0,0 +1,24 @@
...
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Single-qubit instruction to denote a wire cut location."""
+from __future__ import annotations
+
+from qiskit.circuit import Instruction
+
+
+class CutWire(Instruction):
```
garrison, line 19: It _might_ be more appropriate for this to derive from `Gate`, since a `CutWire` represents an identity gate, which is unitary.  Also, it might make sense for it to have a `_define` method, just like `Move` does.

test/cutting/qpd/test_cut_wire_to_move.py
```diff
@@ -0,0 +1,122 @@
...
+
+from pytest import fixture, mark, raises
+from qiskit.circuit import QuantumCircuit
+from circuit_knitting.cutting.qpd.instructions.move import Move
+from circuit_knitting.cutting.qpd.instructions.cut_wire import CutWire
+from circuit_knitting.cutting.qpd.cut_wire_to_move import transform_to_move
+
+
+@fixture
+def circuit1() -> QuantumCircuit:
```
garrison, line 24: It'd be nice to have `resulting_circuit1` come immediately after this, so I don't have to scroll past all the others to do a comparison with my eyes.

circuit_knitting/cutting/qpd/cut_wire_to_move.py
```diff
@@ -0,0 +1,84 @@
...
+    Returns:
+        circuit (QuantumCircuit): new circuit with :class`.move` instructions.
+
+    Raises:
+        ValueError: circuit (QuantumCircuit) contains no :class:`.cut_wire` instructions.
+    """
+
+    new_circuit, mapping = _circuit_structure_mapping(circuit)
+
+    for _, instructions in enumerate(circuit.data):
```
garrison, line 36:
```suggestion
    for instructions in circuit.data:
```

test/cutting/qpd/test_cut_wire_to_move.py
```diff
@@ -0,0 +1,142 @@
...
+
+@mark.parametrize(
+    "sample_circuit, resulting_circuit",
+    [
+        ("circuit1", "resulting_circuit1"),
+        ("circuit2", "resulting_circuit2"),
+        ("circuit3", "resulting_circuit3"),
+    ],
+)
+def test_circuit_registers(request, sample_circuit, resulting_circuit):
```
garrison, line 132: This a a good start for the register tests.  I also think they could be improved by adding a number of test cases with additional complexity.

- if there's more than one register, do the correct register(s) grow in size based on the number of `CutWire` instructions on each one's qubit wires?
- if there are bare qubits in addition to qubits container in register(s), do the additional qubits get added as expected?  (registers should grow, and bare qubits should be added alongside bare qubits).
- if the ordering of qubits is perverse, do things still behave properly?  Think of something like the below example, which contains qubits in registers as well as bare qubits, but the register qubits aren't even in consecutive locations in the input circuit.

```python
from qiskit.circuit import QuantumCircuit, QuantumRegister, Qubit
qc = QuantumCircuit()
reg1 = QuantumRegister(4, "qx")
reg2 = QuantumRegister(4, "qy")
for i in range(4):
    qc.add_bits([reg1[i], reg2[i], Qubit()])
qc.add_register(reg1)
qc.add_register(reg2)
```

(Try that in a jupyter notebook, then inspect `qc.qregs` and `qc.qubits`.)

circuit_knitting/cutting/qpd/cut_wire_to_move.py
```diff
@@ -0,0 +1,82 @@
...
+"""Function to transform a CutWire instruction to a Move instruction."""
+from __future__ import annotations
+
+from itertools import groupby
+from qiskit.circuit import Qubit, QuantumCircuit
+from circuit_knitting.cutting.instructions.move import Move
+
+
+def transform_to_move(circuit: QuantumCircuit) -> QuantumCircuit:
+    """Transform a :class:`.cut_wire` instruction to a :class:`.move` instruction.
```
garrison:
```suggestion
    """Transform a :class:`.CutWire` instruction to a :class:`.Move` instruction.
```

circuit_knitting/cutting/instructions/cut_wire.py
```diff
@@ -0,0 +1,29 @@
...
+class CutWire(Gate):
+    """An instruction for denoting a wire cut location."""
+
+    def __init__(self, label: str | None = None):
+        """Create CutWire instruction."""
+        super().__init__("cut_wire", 1, [], label)
+
+    def _define(self):
+        circuit = QuantumCircuit(1, name=self.name)
+        circuit.i(0)
```
garrison, line 28:
```suggestion
```

I actually think we should remove this line.

From https://qiskit.org/documentation/stubs/qiskit.circuit.library.IGate.html:
> Identity gate corresponds to a single-qubit gate wait cycle, and should not be optimized or unrolled (it is an opaque gate).

so it doesn't actually mean "do nothing", it means to wait for some specified time period, at least in the current version of Qiskit.

circuit_knitting/cutting/cut_wire_to_move.py
```diff
@@ -0,0 +1,82 @@
...
+            new_circuit = new_circuit.compose(
+                other=Move(),
+                qubits=[mapping[gate_index[0]], mapping[gate_index[0]] + 1],
+            )
+            mapping[gate_index[0]] += 1
+        else:
+            new_circuit = new_circuit.compose(
+                other=instructions[0],
+                qubits=[mapping[index] for index in gate_index],
+            )
```
garrison, line 46: Is there a reason you used `compose` rather than `append`?  I am worried because compose creates a new circuit each time (at least the default version, with `inplace=False`), which adds an additional linearly time complexity to this.

https://qiskit.org/documentation/stubs/qiskit.circuit.QuantumCircuit.html#qiskit.circuit.QuantumCircuit.compose

circuit_knitting/cutting/cut_wire_to_move.py
```diff
@@ -0,0 +1,82 @@
...
+
+    Returns:
+        circuit (QuantumCircuit): new circuit with :class`.move` instructions.
+    """
+    new_circuit, mapping = _circuit_structure_mapping(circuit)
+
+    for instructions in circuit.data:
+        gate_index = [circuit.find_bit(qubit).index for qubit in instructions.qubits]
+
+        if instructions in circuit.get_instructions("cut_wire"):
```
garrison, line 35: `circuit.get_instructions("cut_wire")` can probably be called only once, outside the loop, I think.

garrison: or, you could even just look at the current instruction to see if it is a `Move` or not

circuit_knitting/cutting/cut_wire_to_move.py
```diff
@@ -0,0 +1,84 @@
...
+"""Function to transform a :class:`.CutWire` instruction to a :class:`.Move` instruction."""
+from __future__ import annotations
+
+from itertools import groupby
+from qiskit.circuit import Qubit, QuantumCircuit
+from circuit_knitting.cutting.instructions.move import Move
+
+
+def transform_to_move(circuit: QuantumCircuit) -> QuantumCircuit:
+    """Transform a :class:`.CutWire` instruction to a :class:`.Move` instruction.
```
caleb-johnson, line 22:
```suggestion
    """Transform all :class:`.CutWire` instructions in a circuit to :class:`.Move` instructions.
```

circuit_knitting/cutting/cut_wire_to_move.py
```diff
@@ -0,0 +1,84 @@
...
+
+    Args:
+        circuit (QuantumCircuit): original circuit with :class:`.cut_wire` instructions.
```
caleb-johnson, line 25:
```suggestion
        circuit: Original circuit with :class:`.CutWire` instructions
```

circuit_knitting/cutting/cut_wire_to_move.py
```diff
@@ -0,0 +1,84 @@
...
+
+    Returns:
+        circuit (QuantumCircuit): new circuit with :class`.move` instructions.
```
caleb-johnson, line 28:
```suggestion
        circuit: New circuit with :class:`.CutWire` instructions replaced by :class`.Move` instructions
```

circuit_knitting/cutting/cut_wire_to_move.py
```diff
@@ -0,0 +1,84 @@
...
+
+"""Function to transform a :class:`.CutWire` instruction to a :class:`.Move` instruction."""
+from __future__ import annotations
+
+from itertools import groupby
+from qiskit.circuit import Qubit, QuantumCircuit
+from circuit_knitting.cutting.instructions.move import Move
+
+
+def transform_to_move(circuit: QuantumCircuit) -> QuantumCircuit:
```
garrison, line 21: I realized late last week that it would be even better for this function to return a `Move` instruction that has already been wrapped in a `TwoQubitGate`.  So, instead of `new_inst = Move()`, it would be something like `new_inst = TwoQubitQPDGate(QPDBasis.from_gate(Move()), label="cut_move")`.  (We should probably make a shorter way for obtained a wrapped instruction like this eventually, after some discussion, but this is the method that currently works.)

This change would allow the output of this function to be sent directly to `partition_problem` once #350 is merged.

With this in mind, I wonder if something like `place_cut_wires` would be a better name for this function.  (`preprocess_cut_wires` is another potential name that I thought of earlier.)  @caleb-johnson, what do you think?)

garrison: > With this in mind, I wonder if something like `place_cut_wires` would be a better name for this function. (`preprocess_cut_wires` is another potential name that I thought of earlier.) @caleb-johnson, what do you think?)

I think my earlier name might actually be better, since the `CutWire`s are not being placed, they're actually being _re_-placed.  I still find the phrase `preprocess_cut_wires` to be a little bit vague though.  And it's really _processing_ the cut wires -- I called it preprocess only because it's meant to happen early in the workflow, e.g. before `partition_problem`.

garrison: > I realized late last week that it would be even better for this function to return a `Move` instruction that has already been wrapped in a `TwoQubitGate`.

One competing idea to be to merge this function in its current form, since it adds immediate value.  Then, in the next release (v0.4), we could introduce the new function that I suggest (both of these functions could be implemented by calling a common private function, which would contain the guts of the current PR), and we could deprecate this function if it is no longer expected to provide value.  This might be a more straightforward path.

caleb-johnson: We say in the docs we play fast and loose with the terms "cut/partition" to avoid using the term "decompose" when describing the act of setting a circuit up for QPD. I think it's a fair vaguery ;)

garrison: I renamed this function to `transform_cuts_to_moves` in 1200e646e186f2f68cbe955bba39ea1fc44ce0c5.  The additional (future) function, which will return `TwoQubitQPDGate`s, will need a different name anyway.  `cut_wires` might indeed make the most sense for that function; we can decide later.

garrison:
I decided this doesn't need its own release note.  There is already one that says

> The `circuit_knitting.cutting` module now supports wire cutting.  There is a new tutorial [linked] that explains how to use it.

I'd like to merge this soon and then work on a how-to in a subsequent PR.

garrison:
Thanks, @SaashaJoshi!  This will be in our next release (v0.3), which we expect to tag later this week.

PR #328, opened 2023-07-17
Migrate to Docker Compose V2

garrison:
From https://docs.docker.com/compose/migrate/:

> From July 2023 Compose V1 stopped receiving updates. It’s also no longer available in new releases of Docker Desktop.
>
> [...]
>
> Unlike Compose V1, Compose V2 integrates into the Docker CLI platform and the recommended command-line syntax is `docker compose`.

From https://docs.docker.com/compose/compose-file/03-compose-file/:

> The default path for a Compose file is `compose.yaml` (preferred) or `compose.yml` that is placed in the working directory. Compose also supports `docker-compose.yaml` and `docker-compose.yml` for backwards compatibility of earlier versions.

From https://github.com/compose-spec/compose-spec/blob/master/spec.md:

> The top-level `version` property is defined by the Compose Specification for backward compatibility. It is only informative.

hence the removal of `version` from the yaml file.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -38,4 +38,4 @@ This workflow is triggered by a maintainer pushing a tag that represents a relea
 
 ## Docker (`docker.yml`)
 
-This workflow runs periodically (weekly, at the time of writing) to ensure that the [`Dockerfile`](/Dockerfile) and [`docker-compose.yml`](/docker-compose.yml) files at the root of the repository result in a successful build with notebooks that execute without error.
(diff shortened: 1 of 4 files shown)


caleb-johnson:
I guess we should update the docker instructions in INSTALL.md to the more modern syntax (e.g. `docker-compose` --> `docker compose`)

garrison:
> I guess we should update the docker instructions in INSTALL.md to the more modern syntax (e.g. `docker-compose` --> `docker compose`)

I edited `INSTALL.rst`, but do you see a place I missed?

caleb-johnson:
> > I guess we should update the docker instructions in INSTALL.md to the more modern syntax (e.g. `docker-compose` --> `docker compose`)
> 
> I edited `INSTALL.rst`, but do you see a place I missed?

Yes, you did! Sorry bout that, disregard my comment :)

PR #329, opened 2023-07-18
Switch to Qiskit Ecosystem theme

Eric-Arellano:
The main changes are:

1. Gets rid of the top Qiskit nav bar.
2. Adds the project name and Qiskit Ecosystem logo to the left sidebar. (You can disable the logo or replace it with your own)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -59,7 +59,8 @@ pygments_style = "colorful"
 add_module_names = False
 modindex_common_prefix = ["circuit_knitting."]
 
-html_theme = "qiskit"
+html_theme = "qiskit-ecosystem"
+html_title = f"{project} {release}"
 
 # autodoc/autosummary options
 autosummary_generate = True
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -59,7 +59,8 @@
 add_module_names = False
 modindex_common_prefix = ["circuit_knitting."]
 
-html_theme = "qiskit"
+html_theme = "qiskit-ecosystem"
+html_title = f"{project} {release}"
```
Eric-Arellano, line 63: This default to `f"{project} {release} documentation"`, which is redundant and noisy. The value gets used for the left sidebar & page title in the browser.

garrison: The only possible comment I can think of is that `f"{project}"` _might_ be sufficient.  But I don't see anything wrong with including the version, too.

Eric-Arellano: Yeah, for the Sphinx Docs Guide I went with just `project` since versioning doesn't matter there. But I figure most projects do care about versioning so should include it.

PR #333, opened 2023-07-20
Dont batch unless separate samplers are used for each subcircuit

### LINKED ISSUE #330: execute_experiments should not batch subcircuit experiments unless multiple samplers are passed
Batching when only one `Sampler` is passed results in multiple consecutive jobs, and each job contains significant queuing and error mitigation overhead. We shouldn't batch unless the user has set the subcircuits up for different backends or with different sets of options.

caleb-johnson:
Resolves #330


circuit_knitting/cutting/cutting_evaluation.py
```diff
@@ -327,18 +343,14 @@ def _run_experiments_batch(
...
-    else:
-        # We manually build the shape tuple in second arg because it behaves strangely
-        # with QuantumCircuits in some versions. (e.g. passes local pytest but fails in tox env)
-        quasi_dists_reshaped = np.reshape(
-            quasi_dists_flat, (len(subexperiments), len(subexperiments[0]))
-        )
-        num_qpd_bits = np.reshape(
-            num_qpd_bits_flat, (len(subexperiments), len(subexperiments[0]))
-        )
+    quasi_dists_reshaped: list[list[QuasiDistribution]] = [[] for _ in subexperiments]
```
caleb-johnson, line 346: Can't rely on numpy here anymore because there are different subcircuits coming in on the same batch, and those can have different numbers of subexperiments associated with them; therefore, we can't trivially reshape to the input size anymore.

We now have to "stream" things into the data structure, using the input as a guide for how to re-build the structure

garrison:
> If we get this one merged soon, it's probably worth a patch release to 0.2.1 due to the large amount of time saved in cases where the circuit is separated and the user runs on quantum hardware with no reservation

That reminds me: this definitely deserves a release note.

caleb-johnson:
> > If we get this one merged soon, it's probably worth a patch release to 0.2.1 due to the large amount of time saved in cases where the circuit is separated and the user runs on quantum hardware with no reservation

This thought occurred to me before my coffee. We already have features in main, it's too late to consider a bugfix release probably :)

caleb-johnson:
@garrison the change log should now be correct in this PR now, so it is actually reviewable :)

garrison (approved):
Thanks, this looks reasonable, and it appears basically correct as far as I can tell without working out every last detail explicitly about the orderings in these reshaped lists.  In order to raise my confidence to a pretty high bar, I tested this with a modified roundtrip test, and everything works out.  I will submit my modified test in a PR soon, as I think it is worth doing on an ongoing basis.  Thanks!

PR #334, opened 2023-07-21
Remove slow tests that don't add coverage

caleb-johnson:
There is no reason to keep the slow tests that don't any any code coverage or special case coverage. This PR removes tests marked slow which have total overlap with the remainder of the test suite.


garrison:
I think this can be closed in favor of 4960b632d1a560cd44f4c7575d7435f19ddf340a.  (Sorry, I meant for that to be a pull request but accidentally pushed to `main`.)

PR #336, opened 2023-07-26
Improve the instructions regarding pandoc

garrison:
I've also tried to make the developer documentation easier to discover

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -27,6 +27,11 @@ If you use the Circuit Knitting Toolbox in your research, please cite it accordi
 .. literalinclude:: ../CITATION.bib
    :language: bibtex
 
+Developer guide
+---------------
+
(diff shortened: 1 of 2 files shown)


IbrahimShehzad (changes requested):
These improvements are very helpful and I checked that following these revised instructions all the way through led me to the desired result. Even though I understand what you mean, I wonder if its _completely_ accurate to say that ''_pandoc is not available via pip_...'' (see, e.g, https://pypi.org/project/pandoc/, where it appears that you can do a `conda install` followed by a `pip install`). Maybe we say ''it is not sufficient to _just_ run `pip install pandoc`," or something to that effect? I could just be splitting hairs though :)

garrison:
> I wonder if its _completely_ accurate to say that ''_pandoc is not available via pip_...'' (see, e.g, https://pypi.org/project/pandoc/

That is the "Pandoc python library", which depends on but does not provide pandoc, hence the need to install it from some other source first (one option of which is via conda).

IbrahimShehzad (approved):

PR #338, opened 2023-07-27
Make the repository link more obvious from the Sphinx build

garrison:
This PR does two things to make the repository more discoverable from the Sphinx build
- moves the mention of the GitHub repository to the section titled "Developer guide"
- adds a GitHub icon in the footer of each page, which links to the repo, following the instructions at https://pradyunsg.me/furo/customisation/footer/#using-embedded-svgs

#### Screenshot of the footer (see bottom right corner)
furo-github-footer

PR #339, opened 2023-07-27
Add README badge linking to stable documentation

garrison:
This is complementary to #338, but in the reverse direction.  I want to make sure the docs are easily discoverable from github if one does not know to look in the sidebar.  I also added a ruff badge for good measure, but I could be convinced this is unnecessary.  Preview at https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/docs-badge/README.md


caleb-johnson (dismissed):
I'd probably leave ruff out bc I find it hard to believe anyone besides me and you cares too much about our linter ;). Up to you really though. Love the idea of the docs badge

PR #342, opened 2023-07-28
Add sampling overhead table and expand/update cutting explanation

### LINKED ISSUE #293: Add a brief note about "global phase" in the cutting explanatory material
Circuit cutting involves implementing quantum channels that will result in the same results for physically measurable things, but "global phase" is not something that can be physically measured, so it's something we ignore when coming up with equivalences.  We should clearly mention this in the explanatory material.

### LINKED ISSUE #373: Add `Move`/`CutWire` to explanatory material
Continuation of conversation at https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/370#discussion_r1297226937.

garrison:
This adds a sampling overhead table.  ~~I should triple check it.~~  I may push other improvements to the cutting explanation to this branch, too.

EDIT: I've added many other improvements, too.

Fixes #293.
Fixes #373.


garrison:
I will likely keep tweaking this, and I still plan to expand the explanation document a bit (either before or after this PR is merged -- doesn't matter to me), but this is already a substantial contribution and is in a good place, essentially ready for review.

The most convenient way to review it probably is to either build it locally or to download the artifact from the Docs CI run.  (Click on the docs run, then click "Summary", then click "html_docs.")

garrison:
@caleb-johnson and @ibrahim-shehzad: Requesting your review.  Any comments are truly welcome, but just remember this doesn't need to be perfect before we merge; we can always iterate on it later.

docs/circuit_cutting/explanation/index.rst
```diff
@@ -10,39 +10,185 @@ Circuit cutting is a technique to increase the size of circuits we can run on qu
...
+The sampling overhead is the factor by which the overall number of shots must increase for the quasiprobability decomposition to result in the same amount of error, :math:`\epsilon`, as one would get by executing the original circuit.
+The overhead of a cut gate depends on which gate is cut; see the final appendix of [`1 <https://arxiv.org/abs/2205.00016>`__] for details.
+For instance, a single cut CNOT gate incurs a sampling overhead of 9 [`2 <https://arxiv.org/abs/1909.07534>`__,\ `6 <https://arxiv.org/abs/2312.11638>`__].
+A circuit with :math:`n` wire cuts incurs a sampling overhead of O(:math:`16^n`) in the LO setting.  If real-time communication *is* available (i.e., if the hardware supports “dynamic circuits”, also known as the LOCC setting), then the sampling overhead for :math:`n` cut wires may be reduced to O(:math:`4^n`) [`4 <https://arxiv.org/abs/2302.03366>`__].
+However, support for wire cutting with classical communication (LOCC) is not yet supported in CKT (see issue `#264 <https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/issues/264>`__).
```
ibrahim-shehzad, line 34: Really minor point, but ''support... being supported'' sounds slightly awkward.
```Suggestion
However, support for wire cutting with classical communication (LOCC) is not yet available in CKT
```

docs/circuit_cutting/explanation/index.rst
```diff
@@ -10,39 +10,185 @@ Circuit cutting is a technique to increase the size of circuits we can run on qu
...
+As mentioned above, the cost of any simulation based on quasiprobability distribution is an exponential sampling overhead.
+The sampling overhead is the factor by which the overall number of shots must increase for the quasiprobability decomposition to result in the same amount of error, :math:`\epsilon`, as one would get by executing the original circuit.
+The overhead of a cut gate depends on which gate is cut; see the final appendix of [`1 <https://arxiv.org/abs/2205.00016>`__] for details.
+For instance, a single cut CNOT gate incurs a sampling overhead of 9 [`2 <https://arxiv.org/abs/1909.07534>`__,\ `6 <https://arxiv.org/abs/2312.11638>`__].
+A circuit with :math:`n` wire cuts incurs a sampling overhead of O(:math:`16^n`) in the LO setting.  If real-time communication *is* available (i.e., if the hardware supports “dynamic circuits”, also known as the LOCC setting), then the sampling overhead for :math:`n` cut wires may be reduced to O(:math:`4^n`) [`4 <https://arxiv.org/abs/2302.03366>`__].
```
ibrahim-shehzad, line 33: I understand that you would need dynamics circuits to implement LOCC but is it accurate to say that dynamic supports and the LOCC setting are the same thing? Perhaps we could rephrase this as: 

```Suggestion
A circuit with :math:`n` wire cuts incurs a sampling overhead of O(:math:`16^n`) in the LO setting. If real-time communication *is* available (i.e., if the hardware supports “dynamic circuits”, which is needed to implement the LOCC setting), then the sampling overhead for :math:`n` cut wires may be reduced to O(:math:`4^n`) [`4 <https://arxiv.org/abs/2302.03366>`__].
```

garrison: One-way LOCC does not need dynamic circuits.

Two-way LOCC requires dynamic circuits.

I _think_ you can actually get the LOCC wire cutting advantage with only one-way LOCC as long as all the wire cuts move information in the same direction across a circuit separation boundary.

So I should probably rephrase this anyway.

garrison: Fixed in 81da6cfab8fdbd6a72c422b645d8bde7d27aa90f.

docs/circuit_cutting/explanation/index.rst
```diff
@@ -10,39 +10,185 @@ Circuit cutting is a technique to increase the size of circuits we can run on qu
...
 
-To perform circuit cutting, one must partition (“cut”) the graph representing a quantum circuit into smaller pieces, which are then executed on available hardware.  The results of the original circuit must then be reconstructed during post-processing, resulting in the desired quantity (e.g., the expectation value of a joint observable).
+To perform circuit cutting, one must partition (“cut”) the graph representing a quantum circuit into smaller piece(s), which are then executed on available hardware.  The results of the original circuit must then be reconstructed during post-processing, resulting in the desired quantity (e.g., the expectation value of a joint observable).
```
caleb-johnson, line 23: This only covers the cutting-for-width case. I tried to improve it, but I feel like we could remove this paragraph and the rest of the doc remains equally clear.

docs/circuit_cutting/explanation/index.rst
```diff
@@ -10,39 +10,185 @@ Circuit cutting is a technique to increase the size of circuits we can run on qu
...
+There are `three settings <https://research.ibm.com/blog/circuit-knitting-with-classical-communication>`__ to consider for circuit cutting.  The first is where only local operations (LO) [i.e., local *quantum* operations] are available.  The other settings introduce classical communication between the circuit executions, which is known in the quantum information literature as LOCC, for `local operations and classical communication <https://en.wikipedia.org/wiki/LOCC>`__.  The LOCC can be either near-time, one-directional communication between the circuit executions (the second setting), or real-time, bi-directional communication (the third setting).
```
caleb-johnson, line 27: Should we mention here we only support Case 1?

docs/circuit_cutting/explanation/index.rst
```diff
@@ -10,39 +10,185 @@ Circuit cutting is a technique to increase the size of circuits we can run on qu
...
+As mentioned above, the cost of any simulation based on quasiprobability distribution is an exponential sampling overhead.
+The sampling overhead is the factor by which the overall number of shots must increase for the quasiprobability decomposition to result in the same amount of error, :math:`\epsilon`, as one would get by executing the original circuit.
```
caleb-johnson, line 30: Isn't there a sampling error we must incur due to this technique, and the sampling drives the sampling error down to epsilon, which approaches but never reaches 0?

Is it more accurate to say ... "same amount of error as one would get by executing the original circuit, within some tolerance epsilon"?

docs/circuit_cutting/explanation/index.rst
caleb-johnson: I see we mention it below

docs/circuit_cutting/explanation/index.rst
```diff
@@ -10,39 +10,185 @@ Circuit cutting is a technique to increase the size of circuits we can run on qu
...
+| :class:`~qiskit.circuit.library.XXPlusYYGate`, | :math:`(|\theta/4|,|\theta/4|,0)` | :math:`\left[1+4\left|\sin(\theta/2)\right|+2\sin^2(\theta/2)\right]^2` |
+| :class:`~qiskit.circuit.library.XXMinusYYGate` |                                   | (independent of :math:`\beta` parameter)                                |
++------------------------------------------------+-----------------------------------+-------------------------------------------------------------------------+
+| :class:`.Move` (cut wire) without classical    | not applicable                    | :math:`4^2=16`                                                          |
+| communication (i.e., in the LO setting)        |                                   |                                                                         |
++------------------------------------------------+-----------------------------------+-------------------------------------------------------------------------+
 
 Current limitations
 -------------------
 * ``PauliList`` is the only supported observable format until no sooner than CKT v0.7.0.
```
caleb-johnson, line 189: I think we can remove this after merging #523

caleb-johnson (commented):
This is excellent work. Im going to make another pass tomorrow, but I left a few small comments

docs/circuit_cutting/explanation/index.rst
garrison: Yup, I was planning to change this line once that merged.  Fixed in 3c3d23945732addc8327219b04a255d756a454e8.

docs/circuit_cutting/explanation/index.rst
garrison: I rephrased in e7783b8ae12f259bad9c18a976090ca0655a4e8a.  I wanted to have something here so we don't go abruptly from QPD talking about different types of cuts without first saying what cutting is.

But then I realized we essentially cover this same material in the first paragraph on this page.  So, I ended up removing it as you suggested in 2b30437a8c6dd37cde3980be3b01b629e4b5498f.

docs/circuit_cutting/explanation/index.rst
garrison: There is always sampling error because we can only ever take a finite number of shots on real (or even ideal) hardware.  The errors get exaggerated during QPD post-processing by the factor $\gamma$ (see below in the document, "This post-processing boosts the magnitude of each measurement outcome ...").  To keep the error the same, one must multiply the number of shots by $\gamma^2$ to keep the standard error ($\epsilon/\sqrt{n}_{shots}$) constant.

I added one word, but do you think this clarifies anything?

```suggestion
The sampling overhead is the factor by which the overall number of shots must increase for the quasiprobability decomposition to result in the same amount of sampling error, :math:`\epsilon`, as one would get by executing the original circuit.
```

ibrahim-shehzad: >To keep the error the same, one must multiply the number of shots by $\gamma^{2}$
 to keep the standard error...
 
 Based on this, it sounds like it's the standard error $(\epsilon/\sqrt{n})$ that is being kept the same, rather than the sampling error $(\epsilon)$.

docs/circuit_cutting/explanation/index.rst
```diff
@@ -10,39 +10,182 @@ Circuit cutting is a technique to increase the size of circuits we can run on qu
...
 
-To perform circuit cutting, one must partition (“cut”) the graph representing a quantum circuit into smaller pieces, which are then executed on available hardware.  The results of the original circuit must then be reconstructed during post-processing, resulting in the desired quantity (e.g., the expectation value of a joint observable).
+There are two types of cuts: gate cuts and wire cuts.  Gate cuts, also known as "space-like" cuts, exist when the cut goes through a gate operating on two (or more) qubits.  Wire cuts, also known as "time-like" cuts, are direct cuts through a qubit wire, essentially a single-qubit identity gate that has been cut into two pieces.  In CKT, a wire cut is represented by introducing a new qubit into the circuit and moving remaining operations after the cut identity gate to the new qubit; see :ref:`wire cutting as move`, below.
```
ibrahim-shehzad, line 23: Super minor point: I realize that using space-like cuts for gate cuts and time-like cuts for wire cuts is common in the cutting literature but I find this terminology very confusing. This is because if we think of time as flowing horizontally, from left to right in a quantum circuit, a space-like cut should be a ''vertical'' cut (since space-like cuts give you ''constant time slices'') which would be more analogous to a wire cut. Do we need to use the ''space-like/time-like cut'' terminology? Alternatively, I could consider getting rid of my relativity baggage :).

docs/circuit_cutting/explanation/index.rst
ibrahim-shehzad: I see, understood, thanks!

docs/circuit_cutting/explanation/index.rst
```diff
@@ -10,39 +10,182 @@ Circuit cutting is a technique to increase the size of circuits we can run on qu
...
+.. math::
+   :label: eq:qpd
+
+   \mathcal{U} = \sum_i a_i \mathcal{F}_i ,
+
+where :math:`\mathcal{U}` is the channel implementing the desired operation, and each :math:`a_i` is a real coefficient corresponding to a quantum channel, :math:`\mathcal{F}_i`, that is realizable on hardware.
+
+Note that because this is a sum of channels, *not* a sum of unitaries, expectation values can be evaluated efficiently [`2 <https://arxiv.org/abs/1909.07534>`__].
+
+Results equivalent to original unitary channel can be obtained by a post-processing method, given the coefficients :math:`a_i` and each the outcome of each experiment corresponding to the different local (:math:`\mathcal{F}_i`) channels.  This post-processing boosts the magnitude of each measurement outcome by :math:`\sum_i \left| a_i \right|`, resulting in a sampling overhead of that quantity squared [`6 <https://arxiv.org/abs/2312.11638>`__]:
```
ibrahim-shehzad, line 45:
```Suggestion
Results equivalent to original unitary channel can be obtained by a post-processing method, given the coefficients :math:`a_i` and the outcome of each experiment corresponding to the different local (:math:`\mathcal{F}_i`) channels. This post-processing boosts the magnitude of each measurement outcome by :math:`\sum_i \left| a_i \right|`, resulting in a sampling overhead of that quantity squared [`6 <https://arxiv.org/abs/2312.11638>`__]:

```

ibrahim-shehzad: Doesn't a measurement correspond to _a_ _particular_ channel? If so, wouldn't it be more accurate to say that magnitude of each measurement (i.e the measurement corresponding to channel (`\mathcal{F}_i)`) is boosted by `a_{i}`? It's the sum over `a_{i}` that I am not sure about.

garrison: It's not so much that the channel gets boosted; it's that the classical outcome of +1 or -1 (which results from the measurement) gets boosted by $\sum_i \left| a_i \right|$ during reconstruction.  I borrowed this language from the recent paper, cited at the end of the line.

docs/circuit_cutting/explanation/index.rst
```diff
@@ -10,39 +10,182 @@ Circuit cutting is a technique to increase the size of circuits we can run on qu
...
+
+.. math::
+   :label: eq:kak
+
+   U = (V_1 \otimes V_2) \exp \left[ i \left( \theta_x \, X \otimes X + \theta_y \, Y \otimes Y + \theta_z \, Z \otimes Z \right) \right] (V_3 \otimes V_4) ,
+
+where :math:`V_1`, :math:`V_2`, :math:`V_3`, and :math:`V_4` are local, single-qubit operations, and the two-qubit portion of the interaction is parametrized entirely by :math:`\vec{\theta} = (\theta_x, \theta_y, \theta_z)`.  By convention, we have chosen :math:`\vec{\theta}` to be in the "Weyl chamber" restricted by :math:`\pi/4 \geq \theta_x \geq \theta_y \geq | \theta_z | \geq 0` [`6 <https://arxiv.org/abs/2312.11638>`__].
+For more information on the KAK decomposition, see Ref. [`7 <https://arxiv.org/abs/quant-ph/0209120>`__].
+
+The code that generates a cut subexperiments from the KAK decomposition currently follows Ref. [`3 <https://arxiv.org/abs/2006.11174>`__], which is now known to be non-optimal.  A provably optimal method has been presented in Ref. [`6 <https://arxiv.org/abs/2312.11638>`__], but this newer method has not yet been implemented in CKT (see issue `#531 <https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/issues/531>`__).
```
ibrahim-shehzad, line 124: Perhaps: (?)
```suggestion
The code that generates subexperiments from the KAK decomposition currently follows Ref. [`3 <https://arxiv.org/abs/2006.11174>`__], which is now known to be non-optimal.  A provably optimal method has been presented in Ref. [`6 <https://arxiv.org/abs/2312.11638>`__], but this newer method has not yet been implemented in CKT (see issue `#531 <https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/issues/531>`__).
```

ibrahim-shehzad (approved):
I think we may just be able to ignore the contention I raised in [this comment](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/342/files#r1562884021) (since the literature uses this terminology). Overall, this looks good to me, thanks for putting this together.

PR #345, opened 2023-08-01
Fix sphinx build by depending on ipykernel directly

garrison:
This adds a dependency on ipykernel, the lack of which caused CI failures in #343 and #344.   https://github.com/jupyter/jupyter-sphinx/pull/226 is my attempt to "fix" this upstream.


caleb-johnson (approved):
Nice catch and fixes (even upstream ;) )

PR #346, opened 2023-08-04
Tweak the README badges

garrison:
- replaced "Black style" with download count
- added some logos and emojis

[preview here](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/badge-tweaks/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,15 +2,15 @@
 <div align="left">
 
   [![Stability](https://img.shields.io/badge/Stability-alpha-f4d03f.svg)](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/releases)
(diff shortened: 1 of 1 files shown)


garrison:
@caleb-johnson, did I go overkill with the "Docs" and "Platforms" emojis?  What do you think? :thinking:  :slightly_smiling_face: 

That's the only thing I can't decide about.  Otherwise, I think this is a slight improvement.

caleb-johnson:
> @caleb-johnson, did I go overkill with the "Docs" and "Platforms" emojis? What do you think? 🤔 🙂
> 
> That's the only thing I can't decide about. Otherwise, I think this is a slight improvement.

I would never say no to more emojis in the README shields. I want them to be as gaudy as possible lol

PR #347, opened 2023-08-06
Update cutting tutorials

caleb-johnson:
This PR updates the tutorials to reflect some more recent changes.

- Re-print circuits in 2 gate cutting tutorials
    - New visualization for single qubit qpd gates
    - `fold=-1` should be used for long circuits in depth tutorial
- Use np.inf for `num_samples` in all tutorials and add one sentence reinforcing this idea.
- Use .4 for the HWEA parameters in tutorials. I like the target expvals here better -- less chance of an outlier (~30%) relative error


docs/circuit_cutting/tutorials/02_gate_cutting_to_reduce_circuit_depth.ipynb
garrison, line 1: This is _extremely_ minor, but IMHO a `zip` would be more clear than `enumerate` together with `arbitrary_bases[i]`.

garrison:
> * `fold=-1` should be used for long circuits in depth tutorial

Can you elaborate on why you prefer this?

I'm neutral on this, but I support the other changes in this PR.

caleb-johnson:
> > * `fold=-1` should be used for long circuits in depth tutorial
> 
> Can you elaborate on why you prefer this?
> 
> I'm neutral on this, but I support the other changes in this PR.

I think it's a lot easier to reason about a circuit if it only flows left to right in time -- even if that means the individual gates are harder to see (might have to zoom in on notebook).

When the same qubit appears two or more times in the rendering, it just becomes less clear what is what overall IMO.

docs/circuit_cutting/tutorials/02_gate_cutting_to_reduce_circuit_depth.ipynb
garrison, line 1: Can you modify the variable names around the loop as follows?
- `arbitrary_bases` -> `arbitrary_basis_ids`
- `basis` -> `basis_id`

garrison (approved):
Looks great -- thanks!

PR #350, opened 2023-08-10
Fix `partition_problem`'s handling of unlabeled `TwoQubitQPDGate`s

garrison:
I noticed when experimenting with `partition_problem` that it blows up if you give it a circuit that contains an unlabeled `TwoQubitQPDGate`.  In our workflows so far, all `TwoQubitQPDGate`s are created within `partition_problem` itself, as the first thing it does is call `partition_circuit_qubits`, which creates the gates with the labels.  However, it's not strictly necessary for the gates to have labels, and people might want to place some `TwoQubitQPDGates` manually in addition to the ones across the partition labels.  (Also, in the future, we might want to make `partition_labels` optional, at which point it will just partition as well as it can using the `TwoQubitQPDGates` that have been placed.)

I double checked, and the added test indeed fails without the accompanying change.

### DIFF
--- a/circuit_knitting/cutting/cutting_decomposition.py
+++ b/circuit_knitting/cutting/cutting_decomposition.py
@@ -223,7 +223,7 @@ def partition_problem(
     for inst in qpd_circuit.data:
         if isinstance(inst.operation, TwoQubitQPDGate):
             bases.append(inst.operation.basis)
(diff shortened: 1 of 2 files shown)


caleb-johnson (approved):
awesome, nice catch 👍

PR #356, opened 2023-08-14
Update language in cutting tutorials

### LINKED ISSUE #280: Clean up language in cutting tutorials
The variable name `simulated_expvals` in the cutting tutorials might give the impression that the expectation values are from a simulator, when that isn't the intended meaning. `reconstructed_expvals` is a more meaningful name in this case.

Make a sweep through the tutorials and make sure when we use the word "simulate", the meaning is clear.

caleb-johnson:
Resolves #280

Say reconstruct instead of simulate, to avoid confusion between QPD simulation and backend simulators


docs/circuit_cutting/tutorials/01_gate_cutting_to_reduce_circuit_width.ipynb
garrison, line 1: `reconstruct_expectation_values` should instead say `reconstructed_expectation_values`

garrison (approved):
Thanks, I appreciate this.  It looks good to me.

PR #357, opened 2023-08-14
Use multiple samplers sometimes in the roundtrip tests

garrison:
This is my PR promised at https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/333#pullrequestreview-1577702611.

Essentially, I really want the roundtrip tests to regularly test both experiment workflows: the one that uses a single sampler, and the one that uses multiple samplers.  Of course, all sampler(s) used should be `ExactSampler`s, and the expectation values are required to work out exactly, up to floating point error.

This PR achieves this by flipping a coin each time a roundtrip test is called.


caleb-johnson:
Thanks for picking this up, this should have been a >= medium priority issue for a while now :)

caleb-johnson (approved):
Nice, thanks 👍

Issue #363, opened 2023-08-15
Refactor circuit cutting modules to expose backend execution to user


caleb-johnson:
The circuit cutting workflows should expose the usage of the Qiskit primitives to the end user. This will necessitate breaking changes throughout the `circuit_knitting.cutting` modules and the removal of the `cutting.cutting_evaluation` module.

PR #364, opened 2023-08-15
Use the same number of clbits for all subexperiments generated from a given subcircuit

caleb-johnson:
Reasoning about the quasi-dist outputs from various `Sampler` implementations requires knowing the size of the two classical registers added during cutting -- `observable_measurements` and `qpd_measurements`. Making all of the subexperiments generated from a given subcircuit have the same size `qpd_measurements` creg makes post-processing more straightforward.

This change will be useful in the #363 refactor as well.

### DIFF
--- a/circuit_knitting/cutting/qpd/qpd.py
+++ b/circuit_knitting/cutting/qpd/qpd.py
@@ -1135,9 +1135,12 @@ def _decompose_qpd_instructions(
         data_id_offset += 1
         circuit.data.insert(i + data_id_offset, inst2)
 
+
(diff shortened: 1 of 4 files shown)


caleb-johnson:
This is no longer necessary in our workflow, so I'm closing this PR

caleb-johnson:
Working with Runtime team to [pack what we need in Runtime SamplerResult metadata](https://github.ibm.com/IBM-Q-Software/qiskit-ibm-primitives/issues/155). We won't need this PR when that is provided.

PR #365, opened 2023-08-15
Avoid unpacking `PartitionedCuttingProblem` in tutorials & how-tos

garrison:
This modifies our use of `partition_problem` in the documentation to avoid unpacking it like a tuple.  We already know that we plan to break the tuple interface in the v0.4 release cycle.  This prepares for the change, by demonstrating code in our documentation that is resilient to our planned changes.  I'm not sure if `PartitionedCuttingProblem` will end up remaining a `namedtuple` or not; if we add a few more elements to it, it might make more sense for it to be a `dataclass`.  Either way, this code should continue to work on v0.4.  I don't want to _hold_ us to this: if there's some other breaking change we need to make, we should do it.  But this change at least provides some hope that our tutorials from v0.3 might work unmodified on v0.4, too.


caleb-johnson (approved):
Good catch, thanks

PR #367, opened 2023-08-16
Allow `partition_labels` to be determined automatically in `partition_problem`

garrison:
... and introduce `TwoQubitQPDGate.from_instruction`.

I've been working on a PR that introduces a how-to for placing single-qubit wire cuts based on #326.  In the course of completing that, I've realized that it would be _very_ nice to have the feature introduced in the current PR, hence its addition to the 0.3 milestone.  Hopefully I can convince you quickly of its value :slightly_smiling_face:.  I'll open the other PR shortly.  It doesn't strictly depend on this, but (as you will see when I do), it becomes much nicer with this.

### DIFF
--- a/circuit_knitting/cutting/cutting_decomposition.py
+++ b/circuit_knitting/cutting/cutting_decomposition.py
@@ -26,7 +26,7 @@ from qiskit.circuit import (
 from qiskit.quantum_info import PauliList
 
 from ..utils.observable_grouping import observables_restricted_to_subsystem
-from ..utils.transforms import separate_circuit
+from ..utils.transforms import separate_circuit, _partition_labels_from_circuit
 from .qpd.qpd_basis import QPDBasis
 from .qpd.instructions import TwoQubitQPDGate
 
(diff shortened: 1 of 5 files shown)


circuit_knitting/cutting/cutting_decomposition.py
```diff
@@ -26,7 +26,7 @@
 from qiskit.quantum_info import PauliList
 
 from ..utils.observable_grouping import observables_restricted_to_subsystem
-from ..utils.transforms import separate_circuit
+from ..utils.transforms import separate_circuit, _partition_labels_from_circuit
```
caleb-johnson, line 29: Maybe we should expose this function in another PR as well?

caleb-johnson: Since we're in a time crunch, we can push this through and make that function public next release. I'd be fine with that too. Up to you

caleb-johnson:
This is great. I just had one concern about importing the hidden function.

I think maybe we leave it as-is for now and consider where the `partition_labels_from_circuit` function should live and make it public in 0.4?

garrison:
> I think maybe we leave it as-is for now and consider where the partition_labels_from_circuit function should live and make it public in 0.4?

Yeah, I think this is the best plan.  Thank you for the speedy review!

caleb-johnson (approved):
Good ideas! Thanks

PR #368, opened 2023-08-16
Add single-qubit wire cutting how-to and `expand_observables` function

garrison:
I put the how-to and the new `expand_observables` function together in the same PR because the how-to requires the function.  I think it's a decent name, but I'd be open to something better.  (We can always choose a new name later, too.)  This is the PR promised in #367.


garrison:
I actually quite like the name `expand_observables` now that it is used alongside `cut_wires` (#370).

circuit_knitting/cutting/cut_wire_to_move.py
```diff
@@ -82,3 +86,63 @@ def _circuit_structure_mapping(
...
+    original_circuit.num_qubits`` and returns new observables such that
+    ``retval.num_qubits == final_circuit.num_qubits``.
+
+    Args:
+        observables: Observables corresponding to ``original_circuit``
+        original_circuit: Original circuit
+        final_circuit: Final circuit, whose qubits the original ``observables`` should be expanded to.
+
+    Returns:
+        New observables, appropriate for the ``final_circuit``.
```
caleb-johnson, line 122:
```suggestion
        New :math:`N`-qubit observables which are compatible with the :math:`N`-qubit ``final_circuit``
```

circuit_knitting/cutting/cut_wire_to_move.py
```diff
@@ -82,3 +86,63 @@ def _circuit_structure_mapping(
...
+            "The `observables` and `original_circuit` must have the same number "
+            f"of qubits. ({observables.num_qubits} != {original_circuit.num_qubits})"
+        )
+    mapping: list[int] = []
+    for i, qubit in enumerate(original_circuit.qubits):
+        try:
+            idx = final_circuit.find_bit(qubit)[0]
+        except CircuitError as ex:
+            raise ValueError(
+                f"The {i}-th qubit of the `original_circuit` cannot be found "
```
caleb-johnson, line 139: So the original and final circuits need to be comprised of the same `Qubit` instances, is that right?

If that's the case, we could add a sentence here telling users explicitly to build their two circuit using the same qubit instances. Or maybe this sentence is clear enough? What do you think?

caleb-johnson (commented):
Going to take a look at notebooks, but a few small comments here. Looking great

caleb-johnson:
Notes on the how-to:
- Link to arXiv, rather than just text
- If we're going to finish out the workflow, we should move it to tutorials. Otherwise, I think you can cut it off after cell 7

circuit_knitting/cutting/cut_wire_to_move.py
```diff
@@ -14,11 +14,15 @@
 from __future__ import annotations
 
 from itertools import groupby
+import numpy as np
+
```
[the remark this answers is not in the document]
garrison: Thanks, it was actually because of this comment that I realized we should be doing relative import of ckt anyway here.  Fixed in 0d08b59f1f66c464e066909e2f4c332871308783.

caleb-johnson:
> Notes on the how-to:
> 
> * Link to arXiv, rather than just text
> * If we're going to finish out the workflow, we should move it to tutorials. Otherwise, I think you can cut it off after cell 7

I know there is value in showing the rest of the workflow is the same. I'll think more about how to make this more "how-to-y"

circuit_knitting/cutting/cut_wire_to_move.py
garrison: The change in 82f1e6a3e25f5cc0aa1e52f82355979e862cbac7 is subtle, but I think it helps to clarify this.  Since this function is meant to be used with `cut_wires` and `cut_wires` already maintains this condition, I don't think we need to belabor it IMO.

garrison:
> * Link to arXiv, rather than just text

thanks, not sure how I missed this (copy and paste fail)

PR #370, opened 2023-08-17
Use automatic `partition_labels` in new `CutWire` how-to

garrison:
This introduces a `cut_wires` function, as suggested at https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/326#discussion_r1293752734, which is the one remaining piece (beyond #368) that is needed to make it so wire cuts can be specified by `CutWire`s alone, without having also to provide the correct `partition_labels`.

I've opened this as a PR against the branch in #368 so GitHub will show only the changes that are unique to this PR.  I still think, however, that #368 should be merged first and then this should go in after it, since they are each logical, discrete changes.


docs/circuit_cutting/how-tos/how_to_specify_cut_wires.ipynb
garrison, line 1: Need to update the text about the partition labels.

circuit_knitting/cutting/cut_wire_to_move.py
```diff
@@ -13,16 +13,34 @@
 """Function to transform a :class:`.CutWire` instruction to a :class:`.Move` instruction."""
 from __future__ import annotations
 
+from typing import Callable
 from itertools import groupby
 import numpy as np
```
caleb-johnson, line 18: I normally lump imports by Python/External/CKT in lots of other places in the repo. Not a huge deal, but if you're indifferent...

circuit_knitting/cutting/cut_wire_to_move.py
```diff
@@ -13,16 +13,34 @@
...
 from qiskit.quantum_info import PauliList
 from circuit_knitting.cutting.instructions.move import Move
+from circuit_knitting.cutting.qpd.instructions import TwoQubitQPDGate
 
 
-def transform_cuts_to_moves(circuit: QuantumCircuit, /) -> QuantumCircuit:
+def cut_wires(circuit: QuantumCircuit, /) -> QuantumCircuit:
+    """Transform all :class:`.CutWire` instructions in a circuit to :class:`.Move` instructions marked for cutting.
+
+    The returned circuit will have one newly allocated qubit for every :class:`.CutWire` instruction.
```
caleb-johnson, line 30: Should we link to the Piveteau/Sutter LOCC wire cutting paper appendix here? THey do a nice explanation of wire cutting as a 2q identity cut / move operation

caleb-johnson (approved):
This looks really elegant. I like the symmetry between `cut_wires` and `cut_gates`. They both prepare a circuit for a gate cutting workflow. I like that we abstract the difference between gate and wire cutting immediately in our workflow.

Just a small, pedantic comment on imports here and also a suggestion we link to the wire-cutting-as-a-2q-IGate-cut appendix in the Piveteau/Sutter paper.

Thanks!

circuit_knitting/cutting/cut_wire_to_move.py
caleb-johnson: Yes, I plan on making a milestone for tidying up the docs by using more links out to explanation (especially in cutting tutorials. they're very verbose.)

garrison: I think you mean issue, rather than milestone? We can add such an issue to the 0.3.1 milestone.

test/cutting/test_wire_cutting_transforms.py
```diff
@@ -205,9 +207,9 @@ def resulting_circuit4() -> tuple[QuantumCircuit, list[int]]:
         ("circuit4", "resulting_circuit4"),
     ],
 )
-def test_transform_cuts_to_moves(request, sample_circuit, resulting_circuit):
+def test__transform_cuts_to_moves(request, sample_circuit, resulting_circuit):
```
caleb-johnson, line 210: what exactly is the extra underscore here?

caleb-johnson (approved):
Looks good, outside an underscore I couldn't explain

test/cutting/test_wire_cutting_transforms.py
garrison: Ha!  This was not intended.  It came about when I did a find-and-replace to add a leading underscore to `transform_cuts_to_moves`, once I realized that only `cut_wires` is immediately useful.  Maybe we'll find a use for `_transform_cuts_to_moves` one day, in which case I expect we'll make it public.  If not, we might later decide to refactor the tests not to depend on it.

caleb-johnson: I thought you were testing whether I was actually reviewing all your PRs or not lol ;)

PR #371, opened 2023-08-17
Rename `QPDBasis.from_gate()` to `QPDBasis.from_instruction()`

garrison:
This renames `QPDBasis.from_gate()` to `QPDBasis.from_instruction()`.  As of v0.3, the name `from_gate` is no longer appropriate, since the function supports `Move`, which is an `Instruction` but not a `Gate`.  Being that #367 introduced `TwoQubitQPDGate.from_instruction()`, I felt that now is the time to also rename `QPDBasis.from_instruction()` for consistency, so users aren't always confused about which calls it `from_gate` and which calls it `from_instruction`.

I left the argument name as `gate` in some places to reduce the needed size of the diff, but I did mark these places as position-only arguments in the new interface, so that the name can be changed later.

I have also renamed `qpdbasis_from_gate` -> `qpdbasis_from_instruction`, but I left this out of the release note because `qpdbasis_from_gate` was not included in the API refs in v0.2.

### DIFF
--- a/circuit_knitting/cutting/qpd/qpd.py
+++ b/circuit_knitting/cutting/qpd/qpd.py
@@ -546,19 +546,51 @@ def decompose_qpd_instructions(
     return new_qc
 
 
-_qpdbasis_from_gate_funcs: dict[str, Callable[[Instruction], QPDBasis]] = {}
+_qpdbasis_from_instruction_funcs: dict[str, Callable[[Instruction], QPDBasis]] = {}
 
 
-def _register_qpdbasis_from_gate(*args):
+def _register_qpdbasis_from_instruction(*args):
     def g(f):
         for name in args:
-            _qpdbasis_from_gate_funcs[name] = f
+            _qpdbasis_from_instruction_funcs[name] = f
         return f
 
     return g
 
 
-def qpdbasis_from_gate(gate: Instruction) -> QPDBasis:
+@deprecate_func(
+    since="0.3.0",
+    package_name="circuit-knitting-toolbox",
+    removal_timeline="no earlier than v0.4.0",
+    additional_msg=(
+        "This function has been renamed to "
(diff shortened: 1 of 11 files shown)


circuit_knitting/cutting/qpd/qpd.py
```diff
@@ -546,19 +546,51 @@ def decompose_qpd_instructions(
...
-            _qpdbasis_from_gate_funcs[name] = f
+            _qpdbasis_from_instruction_funcs[name] = f
         return f
 
     return g
 
 
-def qpdbasis_from_gate(gate: Instruction) -> QPDBasis:
+@deprecate_func(
+    since="0.3.0",
```
caleb-johnson, line 562: Since we are only supporting minor releases right now, should we include the patch version anymore? I've been doing 0.4 in my PRs (I think).

circuit_knitting/cutting/qpd/qpd_basis.py
```diff
@@ -114,12 +115,20 @@ def overhead(self) -> float:
         return self._kappa**2
 
     @staticmethod
-    def from_gate(gate: Instruction) -> QPDBasis:
+    @deprecate_func(
+        since="0.3.0",
```
caleb-johnson, line 119: same Q as above

caleb-johnson (approved):
Looks great, small question about how we denote versions from here on out. Thanks!

circuit_knitting/cutting/qpd/qpd.py
garrison: Regardless of whether we actually make patch-level releases, I think either way is acceptable.  Qiskit includes a mix of conventions, but uses `since=0.x.0` much more often than `since=0.x`.  Currently, CKT contains one `since="0.3"` and one `since="0.3.0"`.  I think it's fine either way.

PR #377, opened 2023-08-20
Fix double-nesting of cutting tutorials & how-tos in table of contents

garrison:
I noticed that the cutting tutorials & how-tos are double nested in the [table of contents](https://qiskit-extensions.github.io/circuit-knitting-toolbox/#contents).  This currently happens because each header appears redundantly, in `README.rst` and `index.rst`.  This PR fixes the redundancy by removing headers from `index.rst`.  They are worth keeping in `README.rst`, because there they will be shown to users who browse to those directories on github.com.

I built the docs locally, and they look great with this change.


caleb-johnson (approved):
nice catch, thanks

Issue #262, opened 2023-06-16
Filter out duplicate subexperiments before execution


garrison:
The QPD bases we have considered so far each have 6 unique elements in their `maps`.  However, if one looks at only a single qubit (either one), there are actually only 5 unique operations, as two of them are the same `QPDMeasure` operation:

https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/58376677cffdf9e0579b7f2ae3c2b8bd92e13ad0/circuit_knitting/cutting/qpd/qpd.py#L243-L250

For a separable circuit, when it comes time to run the sub-experiments, it's only necessary to run 5 distinct sub-experiments on each partition.  However, right now, 6 subexperiments are run instead.  We already [have a test](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/58376677cffdf9e0579b7f2ae3c2b8bd92e13ad0/test/cutting/qpd/test_qpd.py#L243-L271) to make sure we can recognize these duplicate subexperiments.  We should resolve this issue once we no longer send duplicate experiments to the backends.

### More advanced thoughts regarding wire cutting

A lot of cases of duplicate circuits are straightforward, and we should start with them.

A tricky case, however, is in wire cutting.  The `I` measurement and the `Z` measurement can be made through exactly the same circuit; the only difference is whether the corresponding bit of the qpd register is considered or ignored.

https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/669d9fdfa075ab31cc110be7a75940857c0a249c/circuit_knitting_toolbox/circuit_cutting/qpd/qpd.py#L337
https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/669d9fdfa075ab31cc110be7a75940857c0a249c/circuit_knitting_toolbox/circuit_cutting/qpd/qpd.py#L340

Currently, we always take the parity of all available qpd measurement bits, but if we were to have a bit more control in post-processing, we could do this without running each circuit separately.

garrison:
To do this, we'll need to store the weights in a bit more complicated way (https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/385#discussion_r1310677757) and be also to map between subexperiment index and (sample, observable) index #387.

PR #378, opened 2023-08-21
Add a generate_cutting_experiments function

caleb-johnson:
This should be the first PR to be merged into `main` based on #352 .

This creates a public function `generate_cutting_experiments` which largely mimics the current private function of the same name in `cutting_evaluation`. This function is being moved to `cutting_decomposition`, as subexperiments will be created in that module starting in 0.4.0.

This PR should not be merged prior to 0.3.1, so I am putting it on hold.

The private `_generate_cutting_experiments` will be removed along with the module `cutting_evaluation` in a following PR.

### DIFF
--- a/circuit_knitting/cutting/cutting_decomposition.py
+++ b/circuit_knitting/cutting/cutting_decomposition.py
@@ -481,8 +481,10 @@ def _append_measurement_circuit(
     inplace: bool = False,
 ) -> QuantumCircuit:
     """Append a new classical register and measurement instructions for the given ``CommutingObservableGroup``.
+
     The new register will be named ``"observable_measurements"`` and will be
     the final register in the returned circuit, i.e. ``retval.cregs[-1]``.
+
     Args:
         qc: The quantum circuit
         cog: The commuting observable set for
@@ -493,6 +495,7 @@ def _append_measurement_circuit(
             to have the same number of qubits, and the identity map
             (i.e., ``range(qc.num_qubits)``) is used.
         inplace: Whether to operate on the circuit in place (default: ``False``)
+
     Returns:
         The modified circuit
     """
--- a/circuit_knitting/cutting/cutting_decomposition.py
+++ b/circuit_knitting/cutting/cutting_decomposition.py
@@ -17,18 +17,31 @@ from collections import defaultdict
 from collections.abc import Sequence, Hashable
 from typing import NamedTuple
 
+import numpy as np
 from qiskit.utils import deprecate_func
 from qiskit.circuit import (
     QuantumCircuit,
+    ClassicalRegister,
     CircuitInstruction,
     Barrier,
 )
 from qiskit.quantum_info import PauliList
 
-from ..utils.observable_grouping import observables_restricted_to_subsystem
+from ..utils.iteration import strict_zip
+from ..utils.observable_grouping import (
+    observables_restricted_to_subsystem,
+    ObservableCollection,
+    CommutingObservableGroup,
+)
 from ..utils.transforms import separate_circuit, _partition_labels_from_circuit
-from .qpd.qpd_basis import QPDBasis
-from .qpd.instructions import TwoQubitQPDGate
+from .qpd import (
+    QPDBasis,
+    SingleQubitQPDGate,
+    TwoQubitQPDGate,
+    generate_qpd_weights,
+    decompose_qpd_instructions,
+    WeightType,
+)
 
 
 class PartitionedCuttingProblem(NamedTuple):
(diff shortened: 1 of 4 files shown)


garrison:
> This PR should not be merged prior to 0.3.1, so I am putting it on hold.

I'll just comment on this part, which is: once finalized and approved, we can merge this to `main` anytime, as backport-able changes can be backported to the `stable/0.3` branch, and 0.3.1 will be tagged (_if_ we tag such a release) from this branch too.

circuit_knitting/cutting/cutting_decomposition.py
```diff
@@ -282,3 +295,261 @@ def decompose_observables(
...
+                subcirc_qpd_gate_ids[-1].append([i])
+                subcirc_map_ids[-1].append(decomp_id)
+
+    return subcirc_qpd_gate_ids, subcirc_map_ids
+
+
+def _raise_bad_qpd_gate_labels() -> None:
+    raise ValueError(
+        "BaseQPDGate instances in input circuit(s) should have their "
+        'labels suffixed with "_<cut_#>" so that sibling SingleQubitQPDGate '
```
caleb-johnson, line 449: maybe just "_cut_#" , the brackets may be confusing

releasenotes/notes/generate-experiments-554f5ded3bf1a4dc.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    Added a function, :func:`~circuit_knitting.cutting.generate_cutting_experiments`, which generates the cutting subexperiments to be run on the backend - given an input circuit, observable, and number of samples to draw from the input circuit. For each subexperiment, a quasi-distribution will be generated using Qiskit Sampler primitive. :func:`generate_cutting_experiments` also generates the weights associated with each subexperiment. The quasi-distributions and weights are used during expectation value reconstruction.
```
caleb-johnson, line 4: "samples to draw from the quasi-distribution representing the cuts." maybe? hmm

circuit_knitting/cutting/cutting_decomposition.py
```diff
@@ -282,3 +295,261 @@ def decompose_observables(
...
+        weight and the :class:`WeightType`. Each weight corresponds to one unique sample.
+    Raises:
+        ValueError: ``num_samples`` must either be an integer or infinity.
+        ValueError: :class:`SingleQubitQPDGate` instances must have the cut number
+            appended to the gate label.
+        ValueError: :class:`SingleQubitQPDGate` instances are not allowed in unseparated circuits.
+    """
+    if isinstance(num_samples, float):
+        if num_samples != np.inf:
+            raise ValueError("num_samples must either be an integer or infinity.")
```
garrison, line 346: This is actually more restrictive than the condition in `generate_qpd_weights`.  I think that the condition there should be sufficient here, too.

https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/d933abcc708770594329c22b3bf8ef77ee8aff73/circuit_knitting/cutting/qpd/qpd.py#L333-L334

btw, I phrased it there as `not num_samples >= 1` instead of `num_samples < 1` so that it would also fail if `num_samples` is `NaN`.

I plan to add an explanation section about how this works (#284); essentially, `num_samples` sets a parameter $\delta$ (its reciprocal), which is then the threshold between sampled weights and exact weights.  When `num_samples == np.inf`, the threshold is then zero, so everything gets taken exactly.  Being able to choose a float number allows fine tuning this threshold precisely if one desires.

circuit_knitting/cutting/cutting_decomposition.py
```diff
@@ -282,3 +295,261 @@ def decompose_observables(
...
+
+    If the input circuit and observables are not split into multiple partitions, the output
+    subexperiments will be contained within a 1D array.
+
+    If the input circuit and observables is split into multiple partitions, the output
+    subexperiments will be returned as a dictionary which maps a partition label to to
+    a 1D array containing the subexperiments associated with that partition.
+
+    In both cases, the subexperiment lists are ordered as follows:
+        :math:`[sample_{0}observable_{0}, sample_{0}observable_{1}, ..., sample_{0}observable_{N}, ..., sample_{M}observable_{N}]`
```
garrison, line 319: I think this is fine as a first step, but I would really like to move away from this and instead expose an explicit map that allows looking up each (sample, observable) from the index in this flat list and vice versa.  The reason is because some experiments are actually redundant (#262) or potentially even unnecessary, i.e. if the only observables in that subregion are all Pauli `I`s (in which case those experiments can be skipped altogether).

garrison (commented):
Thanks!

Attached are my comments at the moment, though I am going to have to work with this a bit to tell which of the remaining code (if any) is new, or rather the few details that have been modified.  In general, it's easier to review if the pull requests that change code/interfaces are distinct from the ones that purely _move_ code.  This is fine -- I can definitely work with this -- but ideally it would have been

1. Remove the underscore to make the function public, add docstring(s), and add tests in the location appropriate for the code after the move (first PR)
2. Actually move the code itself, without changing anything other than import lines (second PR)

This would also have avoided the brief period in which the function is duplicated (public & private versions in the repo at once).

>  will be removed along with the module `cutting_evaluation` in a following PR.

I wonder if we can/should keep `execute_experiments` only as a deprecated function?  (and remove it in a subsequent release)

circuit_knitting/cutting/cutting_decomposition.py
caleb-johnson: Ya let's think about this more. If you do the dictionary you're back in a situation where you're either asking the user to set up their experiments for the primitive or you're providing a helper function to do it for them. As it is now, users can pass in the outputs of `decompose` straight into the primitive and get their results, which are the same shape as the input.

There is probably some middle ground we can arrive on if we think on it longer.

caleb-johnson: Deleted a comment bc I misunderstood your statement here. Yes, this sounds like a good idea

caleb-johnson:
Closing in favor of #385

PR #385, opened 2023-08-29
Add a public generate_cutting_experiments function

caleb-johnson:
This should be the first PR to be merged into `main` based on #352 .

This creates a public function `generate_cutting_experiments` which largely mimics the current private function of the same name in `cutting_evaluation`.


caleb-johnson:
@garrison The new, public function actually differs from the private function in the format of the output, so I don't see replacing the old function with the new one being any less convoluted than the previous PR. This PR doesn't move the functions and helpers, so it should be more clear to get through

circuit_knitting/cutting/cutting_evaluation.py
```diff
@@ -238,12 +239,56 @@ def _append_measurement_circuit(
...
+        The weights are always a sequence of length-2 tuples, where each tuple contains the
+        weight and the :class:`WeightType`. Each weight corresponds to one unique sample.
+
+    Raises:
+        ValueError: ``num_samples`` must either be an integer or infinity.
+        ValueError: :class:`SingleQubitQPDGate` instances must have the cut number
+            appended to the gate label.
+        ValueError: :class:`SingleQubitQPDGate` instances are not allowed in unseparated circuits.
+    """
+    if isinstance(num_samples, float):
```
caleb-johnson, line 288: bringing over a comment from #385 . Maybe we need a better check here

caleb-johnson: https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/378/files#r1309144633

circuit_knitting/cutting/cutting_evaluation.py
```diff
@@ -306,13 +367,21 @@ def _generate_cutting_experiments(
...
+                subexperiments_legacy[-1][-1].append(meas_qc)
 
-    return subexperiments, coefficients, sampled_frequencies
+    # If the circuit wasn't separated, return the subexperiments as a list
+    subexperiments_out: list[QuantumCircuit] | dict[
+        str | int, list[QuantumCircuit]
+    ] = subexperiments_dict
+    assert isinstance(subexperiments_out, dict)
+    if len(subexperiments_out.keys()) == 1:
+        subexperiments_out = subexperiments_dict[list(subexperiments_dict.keys())[0]]
```
garrison, line 382: I think it is better to look at the input type of `circuits` here rather than the length of unique labels.  If somebody gave us a `dict`, we should give them a `dict`, too, in return, even if their `dict` contained only one element.  Or, in other words, the return type should depend on the _type_ of the input, not the _value_ of the input (a principle known as "type stability").  We should add a test for this case, too.

```suggestion
    if isinstance(circuits, QuantumCircuit):
        assert len(subexperiments_out.keys()) == 1
        subexperiments_out = list(subexperiments_dict.values())[0]
```

circuit_knitting/cutting/cutting_evaluation.py
```diff
@@ -238,12 +239,56 @@ def _append_measurement_circuit(
...
     observables: PauliList | dict[str | int, PauliList],
-    num_samples: int,
-) -> tuple[list[list[list[QuantumCircuit]]], list[tuple[Any, WeightType]], list[float]]:
-    """Generate all the experiments to run on the backend and their associated coefficients."""
+    num_samples: int | float,
+) -> tuple[
+    list[QuantumCircuit] | dict[str | int, list[QuantumCircuit]],
+    list[tuple[float, WeightType]],
+    list[list[list[QuantumCircuit]]],
+]:
```
garrison, line 250: Eventually, we want this to return only the first two elements, right?  (not `subexperiments_legacy`).

I think it would be pretty straightforward to modify `execute_experiments` to work with `subexperiments_out` instead of `subexperiments_legacy`, wouldn't it?  It would, of course, mean changing a data type in `CuttingExperimentResults` and then also dealing with that changed type in `reconstruct_expectation_values`.  (Without looking, I'm guessing this may be where you took things on your `sampler-out` branch?)

But without making this PR take on too much scope, I suggest the following:

- Restore the leading underscore, to make this modified `_generate_cutting_experiments` (one which returns a 3-tuple) be private.
- Move your new docstring to a `generate_cutting_experiments` right above this one.  After the docstring, this function should simply call `_generate_cutting_experiments`, throw away `subexperiments_legacy`, and then return.
- `execute_experiments` can also call `_generate_cutting_experiments`

This way, the interface exposed by `generate_cutting_experiments` will be stable over time.

---

CODA: Considering my follow-up comment about `weights` below, we might have to break this function's interface eventually, anyway -- ideally before v0.4 but we might not have enough time.  It might be nice to do a little bit more thinking about whether we should break this sooner than later (I'll have to think about it over the next few hours, and I don't want to delay giving some feedback while I do).  At the moment, I'm leaning toward just biting the bullet and changing `weights`, even if we don't have time to finish #262, but I want to think more critically about it.

circuit_knitting/cutting/cutting_evaluation.py
```diff
@@ -393,9 +473,13 @@ def _get_bases_by_partition(
...
         for basis_id in subcirc:
-            decomp_id = int(
-                circuits[i].data[basis_id[0]].operation.label.split("_")[-1]
-            )
+            try:
+                decomp_id = int(
+                    circuits[i].data[basis_id[0]].operation.label.split("_")[-1]
+                )
+            except (AttributeError, ValueError):  # pragma: no cover
+                _raise_bad_qpd_gate_labels()  # pragma: no cover
```
garrison, line 481: Is there really no way to test these lines?

circuit_knitting/cutting/cutting_evaluation.py
```diff
@@ -285,18 +330,34 @@ def _generate_cutting_experiments(
...
-    subexperiments: list[list[list[QuantumCircuit]]] = []
-    coefficients = []
-    sampled_frequencies = []
+    # Generate the output experiments and weights
+    subexperiments_dict: dict[str | int, list[QuantumCircuit]] = defaultdict(list)
+    weights: list[tuple[float, WeightType]] = []
+    for i, (subcircuit, label) in enumerate(
+        strict_zip(subcircuit_list, sorted(subsystem_observables.keys()))
+    ):
+        for z, (map_ids, (redundancy, weight_type)) in enumerate(sorted_samples):
```
garrison, line 339: If we were to flip the nesting order of these loops, we could give rid of the `if i == 0` line, I think.

garrison (commented):
Thanks -- I am still pondering some things, but this should be enough to get you started.

garrison:
BTW, you might wonder at times why I seem attached to `execute_experiments`.  The first reason is to maintain compatibility with existing workflows, but there is also a second (more important) reason: if in the future we want to provide advanced users the ability to do special things with their experiments (e.g., bin them by `weight` and then send multiple batches with different `shots` counts), then we're going to need something that looks a lot like `execute_experiments`, just with some additional knobs to tweak.  Even then, we'll want to expose Qiskit Runtime's Sampler directly in the tutorials so that people gain intuition about how it is used.  The `execute_experiments` function will then be a utility that enables more advanced use of Qiskit Runtime.

circuit_knitting/cutting/cutting_evaluation.py
caleb-johnson: I think I was seeing bad behavior from coverage here, let me get back to you

circuit_knitting/cutting/cutting_evaluation.py
caleb-johnson: I've also been thinking it may actually be possible to support a deprecated workflow for a while without an unreasonable amount of effort. Let's see how it shakes out in these upcoming PRs

circuit_knitting/cutting/cutting_evaluation.py
caleb-johnson: Fixed this by removing the check in one private function, which is called strictly after another private function which makes the same check. The other option was to check that private function itself in a test, which seemed overkill.

Also added one test to test the case in the first private function that should catch the bad labels.

test/cutting/test_cutting_decomposition.py
```diff
@@ -284,3 +287,122 @@ def test_unused_qubits(self):
         )
         assert subcircuits.keys() == {"A", "B"}
         assert subobservables.keys() == {"A", "B"}
+
+    def test_generate_cutting_experiments(self):
```
garrison, line 291: I wonder if `generate_cutting_experiments` should go in its own file, instead of `cutting_decomposition.py`, perhaps a name like `cutting_experiments.py`.  I agree that if, at a high level, we think of our problem as 1. decompose 2. execute 3. reconstruct, then this is part of the "decompose" step.  But it's a bit logically distinct from the existing code in `cutting_decomposition.py`, and it would likely help with our organization to keep it separated.  They're really two distinct modules/steps within the decomposition.

Anyway, the reason I commented on this line is because if we do this, the new tests should be placed in something like `test_cutting_[sub]experiments.py` instead.

caleb-johnson: Ya I'm open to this. I don't think it's a blocker for this PR. When we move the function we'll make a decision in that PR. I'm a bit on the fence right now, but I'd like to hear more of your reasoning.

caleb-johnson: I've moved the test to test_evaluation until we decide on where it all should live. This also cleans up the diff in this PR a bit

circuit_knitting/cutting/cutting_evaluation.py
```diff
@@ -238,12 +239,80 @@ def _append_measurement_circuit(
...
+]:
+    """
+    Generate cutting subexperiments and their associated weights.
+
+    If the input, ``circuits``, is a :class:`QuantumCircuit` instance, the
+    output subexperiments will be contained within a 1D array, and ``observables`` is
+    expected to be a :class:`PauliList` instance.
+
+    If the input circuit and observables are specified by dictionaries with partition labels
+    as keys, the output subexperiments will be returned as a dictionary which maps a
```
garrison, line 258:
```suggestion
    as keys, the output subexperiments will be returned as a dictionary which maps each
```

circuit_knitting/cutting/cutting_evaluation.py
```diff
@@ -238,12 +239,80 @@ def _append_measurement_circuit(
...
+    partition label to to a 1D array containing the subexperiments associated with that partition.
```
garrison, line 259:
```suggestion
    partition label to a 1D array containing the subexperiments associated with that partition.
```

circuit_knitting/cutting/cutting_evaluation.py
```diff
@@ -377,7 +470,18 @@ def _get_mapping_ids_by_partition(
...
+                    decomp_id = int(inst.operation.label.split("_")[-1])
+                except (AttributeError, ValueError):
+                    raise ValueError(
+                        "SingleQubitQPDGate instances in input circuit(s) must have their "
+                        'labels suffixed with "_<id>", where <id> is the index of the gate '
+                        "relative to the other gates belonging to the same cut. For example, "
+                        "a two-qubit gate cut can be represented by two SingleQubitQPDGates -- one "
+                        'labeled "<your_label>_0" and one labeled "<your_label>_1".'
+                        "  This allows SingleQubitQPDGates belonging to the same cut to be "
+                        "sampled together."
```
garrison, line 483:
```suggestion
                        "sampled jointly."
```
i.e., the sampling is from the joint probability distribution (I think this change makes it more clear for people who recognize that term, at least.)

circuit_knitting/cutting/cutting_evaluation.py
```diff
@@ -238,12 +239,80 @@ def _append_measurement_circuit(
...
+    If the input, ``circuits``, is a :class:`QuantumCircuit` instance, the
+    output subexperiments will be contained within a 1D array, and ``observables`` is
+    expected to be a :class:`PauliList` instance.
+
+    If the input circuit and observables are specified by dictionaries with partition labels
+    as keys, the output subexperiments will be returned as a dictionary which maps a
+    partition label to to a 1D array containing the subexperiments associated with that partition.
+
+    In both cases, the subexperiment lists are ordered as follows:
+        :math:`[sample_{0}observable_{0}, sample_{0}observable_{1}, ..., sample_{0}observable_{N}, ..., sample_{M}observable_{N}]`
```
garrison, line 262: I find this a bit more clear:

```suggestion
        :math:`[sample_{0}observable_{0}, ..., sample_{0}observable_{N}, sample_{1}observable_{0}, ..., sample_{M}observable_{N}]`
```

Also, not sure if the ellipses render correctly or funny in Sphinx (I haven't checked), but if I were writing a paper I'd use `\ldots` here instead of `...`.

releasenotes/notes/generate-experiments-2ac773442132c78d.yaml
```diff
@@ -0,0 +1,9 @@
+---
+features:
+  - |
+    Added a module, :mod:`circuit_knitting.cutting.cutting_experiments`, which is intended to hold
+    functions used for generating the quantum experiments needed for circuit cutting. This module
+    will initially hold one function, :func:`circuit_knitting.cutting.cutting_experiments.generate_cutting_experiments`,
```
garrison, line 6:
```suggestion
    will initially hold one function, :func:`~circuit_knitting.cutting.cutting_experiments.generate_cutting_experiments`,
```
(probably don't need to display the whole name here)

test/cutting/test_cutting_experiments.py
```diff
@@ -0,0 +1,149 @@
...
+            assert weights == comp_weights
+            assert len(weights) == len(subexperiments["A"])
+            for circ in subexperiments["A"]:
+                assert isinstance(circ, QuantumCircuit)
+
+        with self.subTest("test bad num_samples"):
+            qc = QuantumCircuit(4)
+            with pytest.raises(ValueError) as e_info:
+                generate_cutting_experiments(qc, PauliList(["ZZZZ"]), 0)
+            assert e_info.value.args[0] == "num_samples must be at least 1."
```
garrison, line 89: would also be nice to test specifically that `np.nan` fails, too.

circuit_knitting/cutting/cutting_experiments.py
```diff
@@ -0,0 +1,75 @@
...
+    output subexperiments will be contained within a 1D array, and ``observables`` is
+    expected to be a :class:`PauliList` instance.
+
+    If the input circuit and observables are specified by dictionaries with partition labels
+    as keys, the output subexperiments will be returned as a dictionary which maps each
+    partition label to a 1D array containing the subexperiments associated with that partition.
+
+    In both cases, the subexperiment lists are ordered as follows:
+
+        :math:`[sample_{0}observable_{0}, \ldots, sample_{0}observable_{N}, \ldots, sample_{M}observable_{N}]`
```
garrison, line 44: This is a rehash of my comment at https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/385#discussion_r1311805133

```suggestion
        :math:`[sample_{0}observable_{0}, \ldots, sample_{0}observable_{N}, sample_{1}observable_{0}, \ldots, sample_{M}observable_{N}]`
```

Not sure if you chose to decline or just didn't notice the one remaining change I suggested.

garrison (commented):
This is looking really good; I don't expect to have any further comments once we resolve the open threads.

circuit_knitting/cutting/cutting_experiments.py
caleb-johnson: > This is a rehash of my comment at [#385 (comment)](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/385#discussion_r1311805133)
> 
> Not sure if you chose to decline or just didn't notice the one remaining change I suggested.

I guess it seemed asymmetric to show one indexing from [0,N] and the other indexing from [0,1,M]. I would either include the 1 in each case (I think that was my original way) or remove it in each case (what I have here). What do you think?

garrison: The reason I chose to do it that way is to show explicitly that (1,0) comes immediately after (0, N).  The way you have written it, (0, N) could (in principle) appear _anywhere_ in the list.

garrison: > The way you have written it, (0, N) could (in principle) appear anywhere in the list.

To follow up, the way you wrote it originally did not have this problem, only the way it appears currently.

circuit_knitting/cutting/cutting_evaluation.py
```diff
@@ -377,7 +416,17 @@ def _get_mapping_ids_by_partition(
...
             if isinstance(inst.operation, SingleQubitQPDGate):
-                decomp_id = int(inst.operation.label.split("_")[-1])
+                try:
+                    decomp_id = int(inst.operation.label.split("_")[-1])
+                except (AttributeError, ValueError):
+                    raise ValueError(
+                        "SingleQubitQPDGate instances in input circuit(s) must have their "
+                        'labels suffixed with "_<id>", where <id> is the index of the cut '
+                        "relative to the other cuts in the circuit. For example, all "
+                        "SingleQubitQPDGates belonging to the same cut, N, should have labels "
```
garrison, line 426: Since this is an example (and not a strict requirement), I prefer "could" rather than "should" here.  Or perhaps there's an even better way to phrase it. ... like "One way to satisfy this is to have ..."

```suggestion
                        "SingleQubitQPDGates belonging to the same cut, N, could have labels "
```
(will also require an update to the test)

caleb-johnson: I'm not sure I understand? Having an integer appended to the end of the label is required here

caleb-johnson: There is an implied "[If using this function], SingleQubitQPDGates *should* have labels ..."

caleb-johnson: Maybe you are right though...maybe the integer requirement is too restrictive. We could handle strings and use dictionaries....

garrison: sorry, the relevant part of this sentence (whose meaning would change with "could") is actually on the next line.  It _needs_ to have the suffix, but it doesn't _need_ to be `formatted as "<gate_label>_N"` (it could be anything, but it's the gate label if you use the other parts of our workflow).

caleb-johnson: Ohhhh I see. I was using `gate_label` generically, but it actually has a specific meaning.

Maybe we can leave "should" and change this to `<your_label>_N`?

garrison: > Maybe we can leave "should" and change this to <your_label>_N?

Yeah, that sounds great.

circuit_knitting/cutting/__init__.py
```diff
@@ -86,7 +87,11 @@
     decompose_gates,
     PartitionedCuttingProblem,
 )
-from .cutting_evaluation import execute_experiments, CuttingExperimentResults
+from .cutting_evaluation import (
+    execute_experiments,
+    CuttingExperimentResults,
+)
```
garrison, line 93: totally up to you, but this could all be put on one line again (and then this line would vanish from the diff)

circuit_knitting/cutting/cutting_experiments.py
```diff
@@ -0,0 +1,75 @@
...
+        If the input circuits is a :class:`QuantumCircuit` instance, the output subexperiments
+        will be a sequence of circuits -- one for every unique sample and observable. If the
+        input circuits are represented as a dictionary keyed by partition labels, the output
+        subexperiments will also be a dictionary keyed by partition labels and containing
+        the subexperiments for each partition.
+        The weights are always a sequence of length-2 tuples, where each tuple contains the
+        weight and the :class:`WeightType`. Each weight corresponds to one unique sample.
+
+    Raises:
+        ValueError: ``num_samples`` must either be at least one.
```
garrison, line 65:
```suggestion
        ValueError: ``num_samples`` must be at least one.
```

circuit_knitting/cutting/cutting_evaluation.py
```diff
@@ -69,15 +70,15 @@ def execute_experiments(
...
+        ValueError: The number of requested samples must be at least one.
         ValueError: The types of ``circuits`` and ``subobservables`` arguments are incompatible.
         ValueError: ``SingleQubitQPDGate``\ s are not supported in unseparable circuits.
         ValueError: The keys for the input dictionaries are not equivalent.
         ValueError: The input circuits may not contain any classical registers or bits.
         ValueError: If multiple samplers are passed, each one must be unique.
     """
     if num_samples <= 0:
-        raise ValueError("The number of requested samples must be positive.")
+        raise ValueError("The number of requested samples must be at least 1.")
```
garrison, line 81: should change the condition on the previous line to `not num_samples >= 1`.

releasenotes/notes/generate-experiments-2ac773442132c78d.yaml
```diff
@@ -0,0 +1,9 @@
+---
+features:
+  - |
+    Added a module, :mod:`circuit_knitting.cutting.cutting_experiments`, which is intended to hold
+    functions used for generating the quantum experiments needed for circuit cutting. This module
+    will initially hold one function, :func:`.generate_cutting_experiments`,
+    which can be used to generate quantum experiments, given an input circuit containing
+    :class:`.BaseQPDGate` instances, some observables, and a number
```
garrison, line 8: I _think_ the circuit must contain `TwoQubitQPDGate`s, specifically.

caleb-johnson: What about wire cutting?

caleb-johnson: Also this includes the case where the circuit is partitioned in a dictionary format, so those have SingleQubitQPDGates

garrison: > Also this includes the case where the circuit is partitioned in a dictionary format, so those have SingleQubitQPDGates

Yes, indeed, makes sense. :)

circuit_knitting/cutting/cutting_evaluation.py
```diff
@@ -306,13 +336,22 @@ def _generate_cutting_experiments(
...
             so = subsystem_observables[label]
             for j, cog in enumerate(so.groups):
                 meas_qc = _append_measurement_circuit(decomp_qc, cog)
-                subexperiments[-1][-1].append(meas_qc)
+                subexperiments_legacy[-1][-1].append(meas_qc)
+
+    # If the input was a single quantum circuit, return the subexperiments as a list
+    subexperiments_out: list[QuantumCircuit] | dict[
+        str | int, list[QuantumCircuit]
+    ] = subexperiments_dict
```
garrison, line 348: `subexperiments_dict` is really a `defaultdict`.  Let's convert it to an actual `dict` before we return it.

```suggestion
    ] = dict(subexperiments_dict)
```

garrison (approved):
Thanks, I am happy with this, modulo my one remaining comment.

garrison (approved):

PR #386, opened 2023-08-30
qiskit version

caleb-johnson:
Development version tests are failing because of version confusion between `qiskit` and `qiskit-terra`. This PR resolves this confusion by using only the meta-package `qiskit` in `pyproject.toml`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,9 +30,9 @@ dependencies = [
     # scipy is currently bounded from above because scipy 1.11 removed the
     # `sym_pos` kwarg from `scipy.linalg.solve`, but pyscf has not yet been updated.
     "scipy>=1.5.2,<1.11",
-    "rustworkx>=0.13.0",
+    "rustworkx>=0.12.0",
     "qiskit-aer>=0.12.0",
-    "qiskit>=0.44.0",
+    "qiskit>=0.43.0",
     "qiskit-nature>=0.6.0",
     "qiskit-ibm-runtime>=0.9.2",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ dependencies = [
     "scipy>=1.5.2,<1.11",
     "rustworkx>=0.12.0",
     "qiskit-aer>=0.12.0",
-    "qiskit-terra>=0.24.0",
+    "qiskit>=0.44.0",
     "qiskit-nature>=0.6.0",
     "qiskit-ibm-runtime>=0.9.2",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,6 +30,7 @@ dependencies = [
     # scipy is currently bounded from above because scipy 1.11 removed the
     # `sym_pos` kwarg from `scipy.linalg.solve`, but pyscf has not yet been updated.
     "scipy>=1.5.2,<1.11",
+    "rustworkx>=0.13.0",
     "qiskit-aer>=0.12.0",
     "qiskit>=0.44.0",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -30,9 +30,8 @@ dependencies = [
     # scipy is currently bounded from above because scipy 1.11 removed the
     # `sym_pos` kwarg from `scipy.linalg.solve`, but pyscf has not yet been updated.
     "scipy>=1.5.2,<1.11",
-    "rustworkx>=0.12.0",
```
garrison, line 33: I think bumping the rustworkx version to 0.13.0 was actually the right call, because we import it directly and therefore should list it as an explicit dependency rather than relying on it being a transitive dependency via qiskit (even though there is essentially no risk of that changing -- the most likely reason it might change, though, is it were ever renamed _again_).  0.13.0 is precisely what Qiskit 0.44.0 depends on, so that's the perfect version.  So I'd say `git revert e17fb62d6d1fccfbf3f560000d02c0e5d2c39319`.

pyproject.toml
```diff
@@ -30,9 +30,8 @@ dependencies = [
...
     "qiskit-aer>=0.12.0",
-    "qiskit-terra>=0.24.0",
+    "qiskit>=0.44.0",
```
garrison, line 34: This actually bumps the version of terra, because Qiskit 0.44 == qiskit-terra 0.25.  I think this is worth doing, because it avoid adding dependencies on qiskit-aer and qiskit-ibmq-provider.  But it does mean we should add a brief release note, and edit the version number in the badge in the README.

garrison (commented):
Thanks for taking care of this so quickly.

pyproject.toml
caleb-johnson: OK sounds good, I do think it's the right call to let the meta-package control everything. I believe this is what the Qiskit devs are hoping clients will do sooner than later

pyproject.toml
caleb-johnson: And yes, we had to bump because `qiskit 0.44` needs `rustworkx 0.13`

pyproject.toml
garrison: That's consistent with the message I was trying to convey.

But actually, I've changed my mind.  Why not just change to `qiskit>=0.43.0`?  Then we don't need a release note, and people installing a newer verson of qiskit (everyone except the "minimum version tests") can still get by without ibmq-provider.  (We can also avoid bumping rustworkx, not that it really matters.)

My earlier concern was that the minimum version tests wouldn't figure this out, but since the metapackage pins the version of terra precisely, it won't be a problem.

garrison (dismissed):
Thanks, I am happy with this as long as tests pass. :crossed_fingers:

garrison:
Actually, before merging, I want to make sure the development tests actually ran as expected.

garrison (changes requested):
I am going to open a PR that builds on this one that attempts to get the development version tests working as expected again.  I think I know how this can work.

Also, I can't figure out how to "revoke" my approval without requesting changes :joy:

caleb-johnson:
> I am going to open a PR that builds on this one that attempts to get the development version tests working as expected again. I think I know how this can work.
> 
> Also, I can't figure out how to "revoke" my approval without requesting changes 😂

Down by the CI checks, you'll see Approvals, click the ` Approver dropdown. You'll see your name and an elipses on the right. Click elipses and dismiss review. I have to look around for ~3 minutes every time I want to do that lol

caleb-johnson:
Fixed in #388

PR #388, opened 2023-08-30
Fix fallout from Qiskit metapackage migration

garrison:
This is an alternative to #386, in an attempt to get the "development version tests" to _actually_ use the development version of Qiskit.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,9 +30,9 @@ dependencies = [
     # scipy is currently bounded from above because scipy 1.11 removed the
     # `sym_pos` kwarg from `scipy.linalg.solve`, but pyscf has not yet been updated.
     "scipy>=1.5.2,<1.11",
-    "rustworkx>=0.13.0",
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -32,6 +32,7 @@ dependencies = [
     "scipy>=1.5.2,<1.11",
     "rustworkx>=0.12.0",
     "qiskit-aer>=0.12.0",
+    "qiskit>=0.43.0",
     "qiskit-terra>=0.24.0",
```
garrison, line 36: If CI works as expected, ~~I will add a comment explaining why we depend on _both_ qiskit and qiskit-terra.~~

Actually, scratch that.  If CI works as expected, I will modify the "development version tests" to add the `qiskit-terra` line for us _only_ during the relevant CI run, so that we can avoid having it in this file.

garrison:
Whew, looks like the development version tests are now working as expected.

PR #391, opened 2023-08-31
Refactor reconstruct_expectation_values

caleb-johnson:
As a follow-up to #385 and in the larger effort associated with #363 , this PR refactors `reconstruct_expectation_values` to accept the  outputs from the Sampler primitives directly (with the assumption that the metadata contains QPD bit information).

This PR will also refactor `execute_experiments` to output a data structure which can also be consumed by `reconstruct_expectation_values`. While this will be a breaking change, it will allow us to prevent an abrupt removal of a core function. Additionally, this breaking change actually results in returning *more* info to the user from `execute_experiments`. The user will now have the entire SamplerResult object(s), including the QPD bit information; whereas, before they only got the output quasi-dists and qpd bit info.


circuit_knitting/cutting/cutting_reconstruction.py
```diff
@@ -79,21 +87,26 @@ def reconstruct_expectation_values(
         label: ObservableCollection(subobservables)
         for label, subobservables in subobservables_by_subsystem.items()
     }
+    sorted_subsystems = sorted(subsystem_observables.keys())  # type: ignore
```
caleb-johnson, line 90: Need to circle back on some of these mypy cheats

circuit_knitting/cutting/cutting_evaluation.py
```diff
@@ -354,55 +350,6 @@ def _generate_cutting_experiments(
     return subexperiments_out, weights, subexperiments_legacy
 
 
-def _run_experiments_batch(
```
caleb-johnson, line 357: No longer needed. Logic is now simplified in `execute_experiments` since we can just pass `SamplerResult` instances out

test/cutting/test_cutting_evaluation.py
```diff
@@ -67,7 +66,10 @@ def test_execute_experiments(self):
             quasi_dists, coefficients = execute_experiments(
```
caleb-johnson, line 66: These "quasi_dists" vars could be renamed to "results" throughout, but I'm trying to keep the diff clean for this PR

circuit_knitting/cutting/cutting_evaluation.py
```diff
@@ -323,25 +316,6 @@ def _generate_cutting_experiments(
                 meas_qc = _append_measurement_circuit(decomp_qc, cog)
                 subexperiments_dict[label].append(meas_qc)
 
-    # Generate legacy subexperiments list
```
caleb-johnson, line 326: We actually don't need the legacy list anymore because we will just pass the outputs of Sampler directly out of `execute_experiments`. The 3D list structure will be removed entirely in this PR

circuit_knitting/cutting/cutting_reconstruction.py
```diff
@@ -26,49 +26,68 @@
...
-        A ``list`` of ``float``\ s, such that each float is a simulated expectation
+        A ``list`` of ``float``\ s, such that each float is an expectation
         value corresponding to the input observable in the same position
 
     Raises:
-        ValueError: The number of unique samples in quasi_dists does not equal the number of coefficients.
+        ValueError: ``observables`` and ``results`` are of incompatible types.
         ValueError: An input observable has a phase not equal to 1.
     """
-    if len(coefficients) != len(quasi_dists):
```
caleb-johnson, line 57: I will probably re-add this check in some form

circuit_knitting/cutting/cutting_evaluation.py
```diff
@@ -246,7 +239,6 @@ def _generate_cutting_experiments(
 ) -> tuple[
     list[QuantumCircuit] | dict[str | int, list[QuantumCircuit]],
     list[tuple[float, WeightType]],
-    list[list[list[QuantumCircuit]]],
```
caleb-johnson, line 249: Since the legacy structure isn't needed anymore, the next PR will move this function to the correct module, `cutting_experiments.py`.

We will leave this in place for this PR, as the focus is getting execute-->reconstruct data structures aligned -- not moving large amounts of code.

Does this sound reasonable, @garrison ?

circuit_knitting/cutting/cutting_reconstruction.py
```diff
@@ -26,49 +26,68 @@
...
+            dictionary mapping partition labels to the observables associated with that partition.
 
     Returns:
-        A ``list`` of ``float``\ s, such that each float is a simulated expectation
+        A ``list`` of ``float``\ s, such that each float is an expectation
         value corresponding to the input observable in the same position
 
     Raises:
-        ValueError: The number of unique samples in quasi_dists does not equal the number of coefficients.
+        ValueError: ``observables`` and ``results`` are of incompatible types.
```
caleb-johnson, line 67: Need to raise an error in this func if users don't set the `num_qpd_bits` field in all their results obj's.

circuit_knitting/cutting/cutting_evaluation.py
```diff
@@ -64,10 +62,8 @@ def execute_experiments(
         samplers: Sampler(s) on which to run the sub-experiments.
 
     Returns:
-        - A 3D list of length-2 tuples holding the quasi-distributions and QPD bit information
-          for each sub-experiment. The shape of the list is: (``num_unique_samples``, ``num_partitions``, ``num_commuting_observ_groups``)
-        - Coefficients corresponding to each unique subexperiment's
-          sampling frequency
+        - A list of :class:`~qiskit.primitives.SamplerResult` instances -- one for each partition.
```
garrison, line 65: It's not a _list_; it's either a `dict` that maps each partition label to a `SamplerResult`, or else a single `SamplerResult`.

circuit_knitting/cutting/cutting_evaluation.py
garrison: Thanks -- and yes, that sounds perfect.

test/cutting/test_cutting_evaluation.py
garrison: `quasi_dists` is a bit nice, though, in the sense that it's less vague than `results`, which might e.g. be a `JobResult` or something if you only pay attention to the name.  (Up to you, though.)

garrison (commented):
Thanks, this is looking pretty great already.  Right now, it's a net loss of 13 lines :tada:.  I will comb through it more carefully shortly.

circuit_knitting/cutting/cutting_evaluation.py
```diff
@@ -41,7 +39,7 @@
 class CuttingExperimentResults(NamedTuple):
     """Circuit cutting subexperiment results and sampling coefficients."""
 
-    quasi_dists: list[list[list[tuple[QuasiDistribution, int]]]]
+    results: SamplerResult | dict[str | int, SamplerResult]
```
garrison, line 42: potentially interesting that the field name is plural (`results`) while the type (`SamplerResult`) is singular.  I wonder if it should just be `result`?  I've noticed similar inconsistencies, though, in other parts of the Qiskit ecosystem, such as the app modules iirc.

circuit_knitting/cutting/cutting_evaluation.py
```diff
@@ -64,10 +62,8 @@ def execute_experiments(
         samplers: Sampler(s) on which to run the sub-experiments.
 
     Returns:
-        - A 3D list of length-2 tuples holding the quasi-distributions and QPD bit information
-          for each sub-experiment. The shape of the list is: (``num_unique_samples``, ``num_partitions``, ``num_commuting_observ_groups``)
-        - Coefficients corresponding to each unique subexperiment's
-          sampling frequency
+        - One :class:`~qiskit.primitives.SamplerResult` instance for each partition.
+        - Coefficients corresponding to each unique subexperiment's sampling frequency
```
garrison, line 66: There's something about the word "frequency" that is confusing to me.  Just a suggestion.

```suggestion
        - Coefficients corresponding to each unique subexperiment's contribution to the reconstructed result
```

circuit_knitting/cutting/cutting_evaluation.py
caleb-johnson: Throughout our cutting workflows we've gone plural with `circuits` and `observables`, since they can be represented as a single instance or a dictionary of instances. My thought was that this applies exactly the same here.

garrison: Yes, I agree with your point now.

caleb-johnson:
> Thanks, this is looking pretty great already. Right now, it's a net loss of 13 lines 🎉. I will comb through it more carefully shortly.

Yes, and most of the added lines are documentation and tests; whereas, most of the removed lines are source code. Hooray

circuit_knitting/cutting/cutting_reconstruction.py
```diff
@@ -26,49 +26,70 @@
...
-            (num_unique_samples, num_partitions, num_commuting_observ_groups)
-        coefficients: A sequence of coefficients, such that each coefficient is associated
-            with one unique sample. The length of ``coefficients`` should equal
-            the length of ``quasi_dists``. Each coefficient is a tuple containing the numerical
-            value and the ``WeightType`` denoting how the value was generated.
+        results: The results from running the cutting subexperiments. If the cut circuit
+            was not partitioned between qubits and run separately, the input should be
+            a :class:`~qiskit.primitives.SamplerResult` instance or a dictionary mapping
+            a single partition to the results. If the circuit was partitioned and its
+            pieces run separately, the input should be a dictionary mapping partition labels
```
garrison, line 41:
```suggestion
            pieces ran separately, this argument should be a dictionary mapping partition labels
```

circuit_knitting/cutting/cutting_reconstruction.py
```diff
@@ -26,49 +26,70 @@
...
     Args:
-        quasi_dists: A 3D sequence of length-2 tuples containing the quasi distributions and
-            QPD bit information from each sub-experiment. Its expected shape is
-            (num_unique_samples, num_partitions, num_commuting_observ_groups)
-        coefficients: A sequence of coefficients, such that each coefficient is associated
-            with one unique sample. The length of ``coefficients`` should equal
-            the length of ``quasi_dists``. Each coefficient is a tuple containing the numerical
-            value and the ``WeightType`` denoting how the value was generated.
+        results: The results from running the cutting subexperiments. If the cut circuit
+            was not partitioned between qubits and run separately, the input should be
```
garrison, line 38:
```suggestion
            was not partitioned between qubits and run separately, this argument should be
```

circuit_knitting/cutting/cutting_reconstruction.py
```diff
@@ -26,49 +26,70 @@
...
+            a :class:`~qiskit.primitives.SamplerResult` instance or a dictionary mapping
+            a single partition to the results. If the circuit was partitioned and its
+            pieces run separately, the input should be a dictionary mapping partition labels
+            to the results from each partition's subexperiments.
+
+            The subexperiment results are expected to be ordered in the same way the subexperiments
+            are ordered in the output of :func:`generate_cutting_experiments` -- one result for every
```
garrison, line 45: I'm not certain that this is necessary, but it shouldn't hurt.

```suggestion
            are ordered in the output of :func:`.generate_cutting_experiments` -- one result for every
```

circuit_knitting/cutting/cutting_reconstruction.py
```diff
@@ -26,49 +26,70 @@
...
+            sample and observable, as shown below. The Qiskit Sampler primitive will return the results
+            in the same order the experiments are submitted, so users who do not use :func:`generate_cutting_experiments`
```
garrison, line 47: Same as above, or we could just check that this displays correctly in Sphinx instead of changing it.

```suggestion
            in the same order the experiments are submitted, so users who do not use :func:`.generate_cutting_experiments`
```

circuit_knitting/cutting/cutting_reconstruction.py
```diff
@@ -26,49 +26,70 @@
...
+            to generate their experiments should take care to format their subexperiments correctly before submitting them
+            to the sampler primitive.
```
garrison, line 49:
```suggestion
            to generate their experiments should take care to order their subexperiments as follows before submitting them
            to the sampler primitive:
```

circuit_knitting/cutting/cutting_reconstruction.py
```diff
@@ -26,49 +26,70 @@
...
+
+            :math:`[sample_{0}observable_{0}, \ldots, sample_{0}observable_{N}, sample_{1}observable_{0}, \ldots, sample_{M}observable_{N}]`
+
+        coefficients: The weights associated with each unique subexperiment. Each weight is a tuple
+            containing the scalar value as well as the :class:`WeightType`, which denotes
+            how the value was generated. The contributions from each subexperiment result to the final
+            expectation value will be scaled by this value.
```
garrison, line 56: I might try to avoid the work "weight" here, since the coefficient can be positive _or_ negative.
```suggestion
        coefficients: A sequence containing the coefficient associated with each unique subexperiment. Each element is a tuple
            containing the coefficient (a ``float``) together with its :class:`.WeightType`, which denotes
            how the value was generated. The contribution from each subexperiment will be multiplied by
            its corresponding coefficient, and the resulting terms will be summed to obtain the reconstructed expectation value.
```

circuit_knitting/cutting/cutting_reconstruction.py
```diff
@@ -79,22 +100,33 @@ def reconstruct_expectation_values(
         label: ObservableCollection(subobservables)
         for label, subobservables in subobservables_by_subsystem.items()
     }
+    sorted_subsystems = sorted(subsystem_observables.keys())
 
-    # Assign each weight's sign and calculate the expectation values for each observable
-    for i, coeff in enumerate(coefficients):
-        sorted_subsystems = sorted(subsystem_observables.keys())  # type: ignore
+    # Reconstruct the expectation values
+    for i in range(len(coefficients)):
```
garrison, line 106:
```suggestion
    for i, coeff in enumerate(coefficients):
```
will enable a line to be dropped ...

circuit_knitting/cutting/cutting_reconstruction.py
```diff
@@ -79,22 +100,33 @@ def reconstruct_expectation_values(
...
         current_expvals = np.ones((len(expvals),))
-        for j, label in enumerate(sorted_subsystems):
+        for label in sorted_subsystems:
             so = subsystem_observables[label]
+            coeff = coefficients[i]
```
garrison, line 110: can be dropped combined with above suggestion.  (or alternatively, just brought out of this level of the loop.)
```suggestion
```

circuit_knitting/cutting/cutting_reconstruction.py
```diff
@@ -79,22 +100,33 @@ def reconstruct_expectation_values(
...
+                                "num_qpd_bits"
+                            ],
+                            cog,
+                            outcome,
+                        )
+                    except KeyError:
+                        raise ValueError(
+                            "The num_qpd_bits field must be set in each subexperiment "
+                            "result metadata dictionary."
+                        )
```
garrison, line 129: My goal here is to have the _minimal_ amount of code in the `try` block, so it can be surgically precise.  If, hypothetically, `_process_outcome` were to raise some unrelated `KeyError`, we don't want to catch that here.  (Not that it will, but we only know that from nonlocal reasoning, which is better avoided.)

Also: I bet black will want to reformat this.

```suggestion
                    try:
                        num_qpd_bits = results_dict[label].metadata[i * len(so.groups) + k][
                            "num_qpd_bits"
                        ]
                    except KeyError:
                        raise ValueError(
                            "The num_qpd_bits field must be set in each subexperiment "
                            "result metadata dictionary."
                        )
                    else:
                        subsystem_expvals[k] += quasi_prob * _process_outcome(
                            num_qpd_bits,
                            cog,
                            outcome,
                        )
```

releasenotes/notes/refactor-evaluate-05fe26e94ff68166.yaml
```diff
@@ -0,0 +1,4 @@
+---
+upgrade:
+  - |
+      The :func:`execute_experiments` function now returns a :class:`~qiskit.primitives.SamplerResult` instance for each circuit partition, rather than the 3D list of quasi-distributions returned previously. The quasi-distributions for each subexperiment can be accessed via the ``quasi_dists`` field of :class:`~qiskit.primitives.SamplerResult`. The number of QPD bits contained in each subexperiment will be included in the ``num_qpd_bits`` field of the ``metadata`` dictionary for each experiment result. The output of this function is still valid as input to :func:`reconstruct_expectation_values`.
```
garrison, line 4: This reads better to me without this s.  Just a suggestion.

```suggestion
      The :func:`execute_experiments` function now returns a :class:`~qiskit.primitives.SamplerResult` instance for each circuit partition, rather than the 3D list of quasi-distributions returned previously. The quasi-distribution for each subexperiment can be accessed via the ``quasi_dists`` field of :class:`~qiskit.primitives.SamplerResult`. The number of QPD bits contained in each subexperiment will be included in the ``num_qpd_bits`` field of the ``metadata`` dictionary for each experiment result. The output of this function is still valid as input to :func:`reconstruct_expectation_values`.
```

garrison: should also check the Sphinx link to make sure a leading period is not needed in ``` :func:`execute_experiments` ```.

releasenotes/notes/refactor-reconstruct-45e00c3df1bdd4ff.yaml
```diff
@@ -0,0 +1,4 @@
+---
+upgrade:
+  - |
+    :func:`reconstruct_expectation_values` now takes a :class:`~qiskit.primitives.SamplerResult` instance or a dictionary mapping partition labels to :class:`~qiskit.primitives.SamplerResult` instances as input to the first positional arg, ``results``. ``results`` replaces the old ``quasi_dists`` positional arg. The :class:`~qiskit.primitives.SamplerResult` instances are expected to contain the number of QPD bits used in each circuit input to the Sampler. This should be specified in the ``num_qpd_bits`` field of the experiment result metadata.
```
garrison, line 4: should check sphinx for ``` :func:`reconstruct_expectation_values` ```, to make sure it works fine without being prefixed with a period

garrison: I built the Sphinx docs locally, and yes, the period is necessary for it to work.

```suggestion
    :func:`.reconstruct_expectation_values` now takes a :class:`~qiskit.primitives.SamplerResult` instance or a dictionary mapping partition labels to :class:`~qiskit.primitives.SamplerResult` instances as input to the first positional arg, ``results``. ``results`` replaces the old ``quasi_dists`` positional arg. The :class:`~qiskit.primitives.SamplerResult` instances are expected to contain the number of QPD bits used in each circuit input to the Sampler. This should be specified in the ``num_qpd_bits`` field of the experiment result metadata.
```

circuit_knitting/cutting/cutting_reconstruction.py
caleb-johnson: > 

I think "the pieces were run separately" is correct. I know this isn't phrased exactly like that, but I believe run is the past participle and should be used here? I'm a little rusty with the grammar, so I could be off :)

circuit_knitting/cutting/cutting_reconstruction.py
caleb-johnson: OK, I've never considered that weight couldn't be negative. I've always just considered it a scalar. We should be consistent though, so this is fine by me

circuit_knitting/cutting/cutting_reconstruction.py
garrison: That sounds even better to me.  I agree with your reasoning.
```suggestion
            pieces were run separately, this argument should be a dictionary mapping partition labels
```

releasenotes/notes/refactor-reconstruct-45e00c3df1bdd4ff.yaml
```diff
@@ -0,0 +1,4 @@
+---
+upgrade:
+  - |
+    :func:`.reconstruct_expectation_values` now takes a :class:`~qiskit.primitives.SamplerResult` instance or a dictionary mapping partition labels to :class:`~qiskit.primitives.SamplerResult` instances as input to the first positional arg, ``results``. ``results`` replaces the old ``quasi_dists`` positional arg. The :class:`~qiskit.primitives.SamplerResult` instances are expected to contain the number of QPD bits used in each circuit input to the Sampler. This should be specified in the ``num_qpd_bits`` field of the experiment result metadata.
```
garrison, line 4: For some reason I prefer not abbreviating argument in the release notes.  This change seems to be consistent with the way most formal documentation is written.

While suggesting this change, though, I realized that these aren't _only_ positional arguments because there is nothing stopping them by being specified by keyword instead.  So I rephrased to make it clear that we are talking about the argument in the first position, not to be confused with https://peps.python.org/pep-0570/.
```suggestion
    :func:`.reconstruct_expectation_values` now takes, as its first argument, a :class:`~qiskit.primitives.SamplerResult` instance or a dictionary mapping partition labels to :class:`~qiskit.primitives.SamplerResult`\ s.  This new ``results`` argument replaces the old ``quasi_dists`` argument. The :class:`~qiskit.primitives.SamplerResult` instances are expected to contain the number of QPD bits used in each circuit input to the Sampler. This should be specified in the ``num_qpd_bits`` field of the experiment result metadata.
```

garrison (approved):
Thanks -- this looks great.  Final tally: net addition of 23 lines, even when counting docstrings, new tests, and the release notes :tada:.

PR #392, opened 2023-09-01
Add a release badge

garrison:
I just can't stop messing with the badges.

The PR adds a "Release" badge, which shows the latest pypi release version and links to the release page on GitHub.

I've also modified the "Python" badge to get the supported versions automatically from pypi.  This will show the versions supported by the latest _released_ of CKT version, rather than by `main`.  However, I think this is probably what users will actually be the most interested in: what versions are supported if I run `pip install`, right now?

Preview at https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/release-badge/README.md


caleb-johnson (approved):
Awesome!

PR #404, opened 2023-09-05
Rework tutorials to use the new workflow

### LINKED ISSUE #351: Call out primitives in cutting tutorials more
- At the very top of each tutorial were we list the three steps of the workflow, mention that “execute” stage uses the Primitives to run circuits on the backends
- Add a little more text to [this description](https://qiskit-extensions.github.io/circuit-knitting-toolbox/circuit_cutting/tutorials/01_gate_cutting_to_reduce_circuit_width.html#Generate-and-run-the-cutting-experiments) that highlights the Primitive/Sampler. Maybe just 1 sentence is enough.
- Add a commented-out line in the corresponding cell showing how to instantiate the Qiskit Runtime Sampler
And just a couple other minor suggestions:
- I think the tutorials are using the old package/directory names. E.g., should be fix from circuit_knitting.cutting import cutting_decomposition
- Could we put the package imports in the cells where they are first used? I find that to be more useful
- qiskit terra is now just qiskit so we should update that in the tutorials

### LINKED ISSUE #363: Refactor circuit cutting modules to expose backend execution to user

caleb-johnson:
Fixes #351 Fixes #363 

Needs #391


docs/circuit_cutting/tutorials/01_gate_cutting_to_reduce_circuit_width.ipynb
garrison, line 1: Let's also update the `SparsePauliOp` timeline in `docs/circuit_cutting/explanation/index.rst`, since it is updated from v0.4.0 to v0.5.0 everywhere except there in this PR currently.

garrison, line 1: - [ ] `generate_cutting_experiments` returns `coefficients`, not "weights" (see #412; a variable should change name in each of these tutorials accordingly)
- [ ] The number of variables could be cut down by using dict comprehensions for the results and jobs (e.g., wouldn't need `sampler_a`, `results_a`, ...).  I can suggest some actual code for this tomorrow, but it's going to look something like https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/7d6f3c98b00cb69f3665dbc4e9ff665e04b81f81/test/cutting/test_cutting_workflows.py#L83-L94
- [ ] For Qiskit Runtime, you first create both jobs, and only then call `.result()` on them.  This is because `result()` will block until the job is done, right?  (And if we call it before submitting the second job, they won't be sent to Runtime in parallel -- or at least that's my understanding, but I have yet to verify this.)  I think we should call this out explicitly in comments in that code block.  Again, I am happy to suggest some wording tomorrow.

caleb-johnson: For your last point (Runtime), I had originally written the tutorials to show it this way (get all jobs queued and then collect all results). I'm realizing now, however, that if you have an open session, you can just queue and retrieve each job individually, as you have front-of-queue privileges. I've updated tutorials to remove collecting of jobs. We'll just get the results directly/consecutively.

garrison:
IMO this PR is also is the last step toward fixing #363.

docs/circuit_cutting/tutorials/01_gate_cutting_to_reduce_circuit_width.ipynb
caleb-johnson: I've addressed your comments here, and your last comment is no longer applicable due to my comment above

garrison, line 1: Not a big deal, but I'm puzzled why you renamed the `circuit` to `qc_0` in the first tutorial.  The second tutorial still has `circuit`.  The third tutorial has `qc_0`, but the reason for that is because there is also a `qc_1` that gets explicitly constructed.

garrison (dismissed):
Thanks, looks great.  The only thing that stood out to me is the `qc_0` in the first tutorial.

docs/circuit_cutting/tutorials/01_gate_cutting_to_reduce_circuit_width.ipynb
garrison, line 1: Blocks like the following would look better to me if even the blank lines started with a `#`.

So instead of 

```python
# from qiskit_ibm_runtime import Session, Options, Sampler

# with Session(backend="ibmq_qasm_simulator") as session:
# Set up Qiskit Runtime Sampler primitives.
# samplers = {
#    label: Sampler(Options(execution={"shots": 2**12})) for label in subexperiments.keys()
# }

# Retrieve results from each subexperiment
# results = {
#    label: sampler.run(subexperiments[label]).result()
#    for label, sampler in samplers.items()
# }

# session.close()
```

I think what looks nicer is 

```python
# from qiskit_ibm_runtime import Session, Options, Sampler
#
# with Session(backend="ibmq_qasm_simulator") as session:
# Set up Qiskit Runtime Sampler primitives.
# samplers = {
#    label: Sampler(Options(execution={"shots": 2**12})) for label in subexperiments.keys()
# }
#
# Retrieve results from each subexperiment
# results = {
#    label: sampler.run(subexperiments[label]).result()
#    for label, sampler in samplers.items()
# }
#
# session.close()
```

caleb-johnson: I see. I use the variable `circuit` in the QPD packing loop and just thought I'd fall in line with your tutorial and change the original circuit to qc_0 to make it more explicit in the final block where we use Aer Estimator to compute the exact expval on the original circuit

garrison: Oh, I see.  Another option is to make one of them `qc`.

garrison (approved):

PR #410, opened 2023-09-09
Make the cutting workflow work with any `Hashable` as label

### LINKED ISSUE #390: Cutting workflow should work with any `Hashable` as labels
Currently, `separate_circuit` takes `partition_labels: Sequence[Hashable]`, but the high-level cutting workflow requires that the partition labels either be all integers or all strings.  This should be pretty simple to fix once the refactor we're targeting for v0.4 is complete; essentially, we'll just need to avoid `sort`ing the labels and refer to them by name always.

garrison:
Fixes #390

- [x] Add a test

I'm leaning against adding a release note, because I don't see any reason to draw attention to this.  This is mostly a conceptual cleanup.

### DIFF
--- a/circuit_knitting/cutting/cutting_decomposition.py
+++ b/circuit_knitting/cutting/cutting_decomposition.py
@@ -34,9 +34,9 @@ from .qpd.instructions import TwoQubitQPDGate
 class PartitionedCuttingProblem(NamedTuple):
     """The result of decomposing and separating a circuit and observable(s)."""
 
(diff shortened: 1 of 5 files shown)


garrison:
> So basically, we have moved entirely to dictionary/single-instance representations throughout our workflow, so there is no longer any need to sort anything. Is that right?

Yes

caleb-johnson (approved):
Awesome, thanks for this. Nice to have this typing discrepancy shored up :)

Issue #411, opened 2023-09-11
Explain the difference between "weights" and "coefficients"


garrison:
... including sampled weights vs. exact weights and sampled coefficients vs. exact coefficients.

PR #412, opened 2023-09-11
`generate_cutting_experiments` returns coefficients, not weights

garrison:
I noticed that `generate_cutting_experiments` really returns _coefficients_, not weights.  While weights (e.g., as returned by `generate_qpd_weights`) are all positive and can be rescaled arbitrarily, coefficients can be either positive or negative and cannot be rescaled arbitrarily.  (Actually, the difference between weights and coefficients deserves some explanation, see #411.)

The misleading jump from "coefficients" to `weights` is in the following snippet:

https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/f33317fde7de9f2c6a5c8916ff301a5388e3eedb/circuit_knitting/cutting/cutting_experiments.py#L144-L145

Interestingly, this was already labeled `coefficients` at the call site in `cutting_experiments.py`.

https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/f33317fde7de9f2c6a5c8916ff301a5388e3eedb/circuit_knitting/cutting/cutting_experiments.py#L144-L145

### DIFF
--- a/circuit_knitting/cutting/cutting_experiments.py
+++ b/circuit_knitting/cutting/cutting_experiments.py
@@ -42,7 +42,7 @@ def generate_cutting_experiments(
     list[tuple[float, WeightType]],
 ]:
     r"""
-    Generate cutting subexperiments and their associated weights.
+    Generate cutting subexperiments and their associated coefficients.
 
     If the input, ``circuits``, is a :class:`QuantumCircuit` instance, the
     output subexperiments will be contained within a 1D array, and ``observables`` is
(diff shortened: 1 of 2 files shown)


caleb-johnson:
Thanks, as I was rewriting some of this, I noticed that the coefficients returned from `decompose` are the actual scalars used during reconstruction. I have never really made a distinction between a weight and a scalar, but your explanation about being able to arbitrarily scale the values makes sense to me :)

caleb-johnson (approved):
Thanks!

PR #414, opened 2023-09-11
Avoid calling `QPDBasis.from_gate` in experiments tests

garrison:
We should instead call `from_instruction`, since `from_gate` was deprecated in #371.

This reduces the number of deprecation warnings shown from running the tests.


caleb-johnson (approved):
Awesome, good catch, thanks!

PR #420, opened 2023-09-14
Configure mergify to perform stable backports

garrison:
This PR configures mergify to perform stable backports for us when we add the "stable backport potential" label to a pull request.  This is essentially the same as [the configuration that qiskit-optimization uses](https://github.com/qiskit-community/qiskit-optimization/blob/main/.mergify.yml), except I removed everything regarding "automerge" and the merge queue.  We seem to get by just fine without a merge queue, and if we _do_ decide to use one, we might as well use GitHub's.

Until now, I've been running `git cherry-pick` manually to backport things to the `stable/*` branches, but using mergify instead will provide mode transparency and opportunities to get involved in the process -- and enforce that tests actually pass before changes are committed to the stable branch.

One potential hiccup: I'm actually not sure if Mergify is enabled in Qiskit-Extensions, so we may need an admin to enable it before it will actually work.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,8 @@
+pull_request_rules:
+  - name: backport
+    conditions:
+      - label=stable backport potential
+    actions:
+      backport:
+        branches:
+          - stable/0.4


mtreinish:
Mergify is configured for qiskit-extensions, but only on select repositories that opt-in to using it. I've enabled it for circuit-knitting-toolbox because of this PR. So as soon as this merges you should be good to go.

garrison:
> Mergify is configured for qiskit-extensions, but only on select repositories that opt-in to using it. I've enabled it for circuit-knitting-toolbox because of this PR. So as soon as this merges you should be good to go.

Thank you, @mtreinish!

caleb-johnson (approved):
Thanks, LGTM!

.mergify.yml
```diff
@@ -0,0 +1,8 @@
+pull_request_rules:
+  - name: backport
+    conditions:
+      - label=stable backport potential
+    actions:
+      backport:
+        branches:
+          - stable/0.4
```
caleb-johnson, line 8: Is this something that will need to be updated manually each release?

caleb-johnson, line 8: Is there some way we can automate the version here so we don't have to mess with it?

We use a python package in the docs config [here](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/105821a9482bea60dd79d64dea7d786cc7e59903/docs/conf.py#L33). I don't have any ideas ready for this yml config file, just thought I'd ask

garrison: I think no matter what, it's going to need to be hard-coded here and updated.  That's just the nature of yaml.

In principle, though, we might be able to _automate_ the bumping of version numbers using a tool like Commitizen.  But I'm not sure how it would deal with the patchlevel only being needed in certain locations, e.g. some places we want `0.4.0` and others (such as here) we just want `0.4`.

PR #421, opened 2023-09-27
Use new `Session.from_id()` method in qiskit-ibm-runtime

garrison:
This removes the places where we modify the private member `session._session_id`, which is possible now that https://github.com/Qiskit/qiskit-ibm-runtime/pull/1101/ has been merged.  We'll have to wait for it to be included in a release (and at that time bump the required qiskit-ibm-runtime version number), hence the "on hold" label.


garrison:
We should also test it in a notebook before merging, since these lines are not picked up by CI.

PR #423, opened 2023-09-29
Remove upper version bound on scipy

garrison:
https://github.com/pyscf/pyscf/issues/1783 has now been resolved in a pyscf release, so we no longer have a restriction on scipy.


caleb-johnson (approved):
woohoo!

PR #425, opened 2023-09-29
Use parameterless gates where possible

garrison:
Now that Qiskit [supports singleton gates](https://github.com/Qiskit/qiskit/pull/10314) for non-controlled, parameterless gates, we might as well use parameterless gates in each place possible to take advantage of this.


caleb-johnson (approved):
Nice LGTM 👍

Issue #422, opened 2023-09-27
subexperiments associated with the identity Pauli observable should not be passed to Sampler


caleb-johnson:
We do not need to run circuits associated with the Identity observable, as the measurement readout is trivial.

Now, the cutting workflow in the tutorials will fail ungracefully when the subexperiments associated with the I observable are passed to the Sampler, as they do not contain any end-of-circuit measurements. Passing these circuits results in:

```
ValueError: The 0-th circuit does not have any classical bit. Sampler requires classical bits, plus measurements on the desired qubits.
```

How do we handle this? Do we only return subexperiments which need actual measurement?

garrison:
The way a user can hack around it without requiring any changes to CKT is to add a Pauli Z observable somewhere in the subsystem that is currently throwing an error with all identities.

But even better would be not to run these experiments at all.  I've had this in mind as one of the special cases at https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/issues/262, but it looks like I failed to ever explicitly mention this there.  I _did_ mention it at https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/378#discussion_r1309157716, and I think the follow up at #387 is relevant to fixing this.

I will also think if there's something we can do that doesn't involve breaking interfaces, which #387 would, technically.  We could always write a helper function that elides calling the Sampler if there are no measurements, but then we're back to where we started: hiding the Sampler from the user.  Alternatively, perhaps the sampler could be modified to have a flag that elides execution (rather than errors) if there are no measurements, but it would take time to get that integrated into all the Sampler implementations.

garrison:
> The way a user can hack around it without requiring any changes to CKT is to add a Pauli Z observable somewhere in the subsystem that is currently throwing an error with all identities.

Actually, automating this hack is probably the best way for us to "quick fix" this, too.  We can modify `generate_cutting_experiments` to always have at least one measurement in each circuit, and this can be done in 0.4.1, without breaking anything.

The "better" fix is going to be to skip the circuit evaluation altogether, which will probably involve tweaking an interface and is related to the issues I linked above.

PR #426, opened 2023-09-29
Provide a workaround to #422, Sampler failing when no measurements

garrison:
Temporary workaround to #422.

Needs
- [x] a test


IbrahimShehzad:
I added a test, @garrison, in `test_cutting_roundtrip.py`, called `test_sampler_fail`.

circuit_knitting/cutting/cutting_experiments.py
```diff
@@ -276,14 +276,21 @@ def _append_measurement_circuit(
     if not inplace:
         qc = qc.copy()
 
+    # If the circuit has no measurements, the Sampler will fail.  So, we
+    # measure one qubit as a temporary workaround to
+    # https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/issues/422
+    pauli_indices = cog.pauli_indices
+    if not pauli_indices:
```
caleb-johnson, line 283: What can `pauli_indices` be here? Should we just explicitly state what we expect? (e.g. `if pauli_indices == []`)

Or we could just assert `pauli_indices == []` inside the if?

caleb-johnson (commented):
Will these circuits with 1 measurement be ignored during reconstruction, since they don't actually have non-trivial observables associated with them?

circuit_knitting/cutting/cutting_experiments.py
garrison: > What can pauli_indices be here? Should we just explicitly state what we expect? (e.g. if pauli_indices == [])

It's a sequence of ints.  It contains the qubit indices that need to be measured.

The choice of style here is influenced by [PEP 8](https://peps.python.org/pep-0008/):

> For sequences, (strings, lists, tuples), use the fact that empty sequences are false:
> ```python
> # Correct:
> if not seq:
> if seq:
> ```
> ```python
> # Wrong:
> if len(seq):
> if not len(seq):
> ```
The "nice" thing about doing it this way is it doesn't matter what type the sequence is.  For instance, if it were to be a tuple of indices, then it will still behave correctly with the current code, but it would not if we were to compare with `[]` because `() != []`.

garrison:
>Will these circuits with 1 measurement be ignored during reconstruction, since they don't actually have non-trivial observables associated with them?

Yes

circuit_knitting/cutting/cutting_experiments.py
caleb-johnson: nice, thanks for explanation

caleb-johnson (approved):
LGTM 👍 

Thanks @IbrahimShehzad and @garrison !

circuit_knitting/cutting/cutting_experiments.py
IbrahimShehzad: Wow, this is more profound than I thought. Love it!

garrison:
I am leaning toward waiting first until #434 is merged and then updating this branch accordingly.

test/cutting/test_cutting_roundtrip.py
```diff
@@ -179,3 +180,28 @@ def test_cutting_exact_reconstruction(example_circuit):
     logger.info("Max error: %f", np.max(np.abs(exact_expvals - simulated_expvals)))
 
     assert np.allclose(exact_expvals, simulated_expvals, atol=1e-8)
+
+
+def test_sampler_fail(example_circuit):
+    """This test checks if the sampler throws an error if you pass it a subscircuit with no measurements. Tests temporary workaround to Issue #422.
```
garrison, line 186:
```suggestion
def test_sampler_with_identity_subobservable(example_circuit):
    """This test ensures that the sampler does not throw an error if you pass it a subcircuit with no observable measurements.
    
    Tests temporary workaround to Issue #422.
```

I think this wording is more clear, as it focuses more on what is supposed to happen rather than the _incorrect_ behavior.

test/cutting/test_cutting_roundtrip.py
```diff
@@ -179,3 +180,28 @@ def test_cutting_exact_reconstruction(example_circuit):
...
+
+    This test passes if no exceptions are raised.
+
+    """
+
+    qc = example_circuit
+    observable_to_test = PauliList(
+        ["IIZ"]
+    )  # Without the workaround to Issue #422, this throws a Sampler error.
```
garrison, line 195:
```suggestion
    )  # Without the workaround to Issue #422, this observable causes a Sampler error.
```
(because the error does not get thrown on _this_ line)

Ibrahim-Shehzad (commented):
Everything now looks good to me. Thanks.

Issue #406, opened 2023-09-07
Forging is not yet compatible with Qiskit Nature 0.7 and higher


garrison:
I haven't figured out exactly why yet, but the merging of https://github.com/qiskit-community/qiskit-nature/pull/1248 broke our "development version tests."

```
=================================== FAILURES ===================================
_ TestEntanglementForgingGroundStateSolver.test_entanglement_forging_vqe_hydrogen _

self = <test.forging.test_entanglement_forging_ground_state_solver.TestEntanglementForgingGroundStateSolver testMethod=test_entanglement_forging_vqe_hydrogen>

    @unittest.skipIf(not pyscf_available, "pyscf is not installed")
    def test_entanglement_forging_vqe_hydrogen(self):
        """Test of applying Entanglement Forged Solver to compute the energy of a H2 molecule."""
        # Set up the ElectronicStructureProblem
        driver = PySCFDriver(
            atom="H .0 .0 .0; H .0 .0 0.735",
            unit=DistanceUnit.ANGSTROM,
            charge=0,
            spin=0,
            basis="sto3g",
        )
        driver.run()
        problem = driver.to_problem(basis=ElectronicBasis.AO)
        qcschema = driver.to_qcschema()
        mo_coeff = get_ao_to_mo_from_qcschema(qcschema).coefficients.alpha["+-"]
    
        # Specify the ansatz and bitstrings
        ansatz = EntanglementForgingAnsatz(
            circuit_u=TwoLocal(2, [], "cry", [[0, 1], [1, 0]], reps=1),
            bitstrings_u=[(1, 0), (0, 1)],
        )
    
        # Set up the entanglement forging vqe object
        solver = EntanglementForgingGroundStateSolver(
            ansatz=ansatz,
            optimizer=SPSA(maxiter=0),
            initial_point=[0.0, np.pi / 2],
            mo_coeff=mo_coeff,
        )
    
        # Solve for the ground state energy
>       results = solver.solve(problem)

test/forging/test_entanglement_forging_ground_state_solver.py:125: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
circuit_knitting/forging/entanglement_forging_ground_state_solver.py:299: in solve
    hamiltonian_terms = self.get_qubit_operators(problem)
[...]

PR #429, opened 2023-10-02
Add support for Python 3.12

garrison:
This is very premature (UPDATE: no longer premature, see https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/429#issuecomment-1832735524 but CI now passes), but it provides us a branch to test against Python 3.12 as dependencies catch up.  I "un-did" some of the changes in #325 here, as cplex is likely to be the last dependency to support Python 3.12.  (Then again, I _could_ be surprised, especially if some other dependencies find it difficult to upgrade.)


garrison:
I _think_ the current error may resolve when https://github.com/symengine/symengine-wheels/issues/11 is fixed.

See also Qiskit's Python 3.12 tracking issue: https://github.com/Qiskit/qiskit/issues/10887

Here's what I see why I try to run `tox -e py312` locally on this branch:

```
Collecting symengine<0.10,>=0.9 (from qiskit-terra>=0.21.0->qiskit-aer>=0.12.0)
  Using cached symengine-0.9.2.tar.gz (85 kB)
  Preparing metadata (setup.py) ... error

  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [24 lines of output]
      /home/garrison/serverless/circuit-knitting-toolbox2/.tox/py312/lib/python3.12/site-packages/setuptools/__init__.py:80: _DeprecatedInstaller: setuptools.installer and fetch_build_eggs are deprecated.
      !!
      
              ********************************************************************************
              Requirements should be satisfied by a PEP 517 installer.
              If you are using pip, you can try `pip install --use-pep517`.
              ********************************************************************************
      
      !!
        dist.fetch_build_eggs(dist.setup_requires)
      error: Multiple top-level packages discovered in a flat-layout: ['cmake', 'symengine'].
      
      To avoid accidental inclusion of unwanted files or directories,
      setuptools will not proceed with this build.
      
      If you are trying to create a single distribution with multiple packages
      on purpose, you should not rely on automatic discovery.
      Instead, consider the following options:
      
      1. set up custom discovery (`find` directive with `include` or `exclude`)
      2. use a `src-layout`
      3. explicitly set `py_modules` or `packages` with a list of names
      
      To find more information, look for "package discovery" on setuptools docs.
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.

note: This is an issue with the package mentioned above, not pip.
hint: See above for details.
```

garrison:
It looks like we will need to wait for a version of Qiskit to be released that removes the pin `symengine>=0.9, <0.10`, as there are only Python 3.12 wheels for symengine 0.11.0 (https://github.com/symengine/symengine-wheels/pull/12).  Note that the Qiskit PR that adds support for Python 3.12 changes this pin to be `symengine>=0.9,!=0.10.0`, so as soon as this PR makes it into a released version of Qiskit (and as soon as we are compatible with this version of Qiskit -- see https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/issues/406#issuecomment-1807252032), then we should be able to support Python 3.12 too.  It would be nice if Qiskit were to backport their PR and include it in a 0.45.X release; then we wouldn't have to wait for (compatibility with) 1.0 to have Python 3.12 support.

mtreinish:
My intent is to backport https://github.com/Qiskit/qiskit/pull/11262 once it's working (it's failing rightnow because subunit, a test dependency, is emitting deprecation warnings on import with 3.12). We typically try to include support for new python versions in a release ASAP.

garrison:
Quantum Serverless (really Ray) does not yet support Python 3.12, so I disabled its install in the notebook tests too.  Once QS and cplex catch up, we can remove the special cases for Python 3.12.

garrison:
Ref https://github.com/ray-project/ray/issues/40211#issuecomment-1894179458

pyproject.toml
```diff
@@ -20,6 +20,7 @@ classifiers = [
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
```
caleb-johnson, line 23: So we report that we support it in the package description, but you do not want to add a README shield?

I wonder if we should not report that we support 3.12 until CPLEX catches up or we remove it in favor of a replacement cut-finding package

garrison: > So we report that we support it in the package description, but you do not want to add a README shield?

Actually, the README shield is generated these days by querying the latest released version on pypi, so this PR (plus a release) will result in the shield being updated:

https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/a666e696486b58798084b231eb4ccc76640b5a64/README.md#L7

> I wonder if we should not report that we support 3.12 at all until CPLEX catches up or we remove it in favor of a replacement cut-finding package

I think it's fine to say we support it in terms of the Trove classifiers, because we do.  Partial functionality works, and full functionality will work once there are cplex wheels.  But yeah, reasonable people could have different takes on this.  I do think it's worth pointing out that one should currently expect only partial functionality under Python 3.12, and this is what the note in INSTALL.rst is for.

Long term it'd of course be nice to remove the current cplex stuff in favor of #471.

garrison:
CI should pass once a release is made that contains https://github.com/Qiskit/qiskit-ibm-runtime/pull/1449 (i.e., after the next qiskit-ibm-runtime release).

garrison:
qiskit-ibm-runtime 0.21 has been released, and CI now passes for this PR.  The status indicator is yellow, though, because github expects to see a Python 3.11 check but sees a Python 3.12 check instead.  @caleb-johnson, if you are good with merging, I can modify the expected check thru the github admin and then merge this.

caleb-johnson:
> qiskit-ibm-runtime 0.21 has been released, and CI now passes for this PR. The status indicator is yellow, though, because github expects to see a Python 3.11 check but sees a Python 3.12 check instead. @caleb-johnson, if you are good with merging, I can modify the expected check thru the github admin and then merge this.

Thanks, I've been keeping an eye on this PR. LGTM

PR #430, opened 2023-10-03
Don't overcount expanded sum terms in documentation

garrison:
Since these sums start with 0, they should end with $N-1$; else, there are $N+1$ terms.  Even though we don't explicitly define `N` or `M`, it is most natural for `N` to represent the number of observables or Cholesky operators, rather than have it be one off from that number.


IbrahimShehzad (approved):
Looks good to me.

PR #431, opened 2023-10-03
Use single gate (`SXGate`) to rotate from Z to Y measurement basis

garrison:
While reading the OpenQASM 3 paper (page 24 of https://arxiv.org/abs/2104.14722v2), I noticed that there is actually a way to rotate to the Y basis to the Z basis in a single parameterless gate.  We might as well use that (most efficient) method here before it gets repeated many times after we generate many subexperiments.


caleb-johnson (approved):
Sweet, nice catch

PR #432, opened 2023-10-04
Update index.rst

MonitSharma:


garrison:
> Circuit cutting is a technique to increase the size of circuits we can run on quantum hardware

I can see why this wording is confusing.  Indeed, circuit cutting _decreases_ the size of the (transpiled) circuit, but the result of that is to run a larger circuit than would otherwise be possible.

MonitSharma:
Yeah . I saw that the moment i sent the PR. That's why I closed it so fast.
Thanks for replying

PR #434, opened 2023-10-04
Swap register order, removing need to pass `num_qpd_bits`

garrison:
This swaps the order of the registers such that `qpd_measurements` now comes last.  The motivation here is two-fold:
1. To be consistent with the natural register order when reconstructing probability distributions (#428)
2. To remove the need to pass around `num_qpd_bits`

### DIFF
--- a/circuit_knitting/cutting/cutting_experiments.py
+++ b/circuit_knitting/cutting/cutting_experiments.py
@@ -148,12 +148,13 @@ def generate_cutting_experiments(
             subcircuit = subcircuit_dict[label]
             if is_separated:
                 map_ids_tmp = tuple(map_ids[j] for j in subcirc_map_ids[label])
-            decomp_qc = decompose_qpd_instructions(
-                subcircuit, subcirc_qpd_gate_ids[label], map_ids_tmp
-            )
             for j, cog in enumerate(so.groups):
-                meas_qc = _append_measurement_circuit(decomp_qc, cog)
-                subexperiments_dict[label].append(meas_qc)
+                new_qc = _append_measurement_register(subcircuit, cog)
+                decompose_qpd_instructions(
+                    new_qc, subcirc_qpd_gate_ids[label], map_ids_tmp, inplace=True
+                )
+                _append_measurement_circuit(new_qc, cog, inplace=True)
+                subexperiments_dict[label].append(new_qc)
 
     # If the input was a single quantum circuit, return the subexperiments as a list
     subexperiments_out: list[QuantumCircuit] | dict[
@@ -231,6 +232,35 @@ def _get_bases(circuit: QuantumCircuit) -> tuple[list[QPDBasis], list[list[int]]
     return bases, qpd_gate_ids
 
 
+def _append_measurement_register(
+    qc: QuantumCircuit,
+    cog: CommutingObservableGroup,
+    /,
+    *,
+    inplace: bool = False,
+):
+    """Append a new classical register for the given ``CommutingObservableGroup``.
+
+    The new register will be named ``"observable_measurements"`` and will be
+    the final register in the returned circuit, i.e. ``retval.cregs[-1]``.
+
+    Args:
+        qc: The quantum circuit
+        cog: The commuting observable set for which to construct measurements
+        inplace: Whether to operate on the circuit in place (default: ``False``)
+
+    Returns:
+        The modified circuit
+    """
+    if not inplace:
+        qc = qc.copy()
+
+    obs_creg = ClassicalRegister(len(cog.pauli_indices), name="observable_measurements")
+    qc.add_register(obs_creg)
+
+    return qc
+
+
 def _append_measurement_circuit(
     qc: QuantumCircuit,
     cog: CommutingObservableGroup,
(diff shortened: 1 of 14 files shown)


circuit_knitting/cutting/cutting_experiments.py
```diff
@@ -273,19 +305,29 @@ def _append_measurement_circuit(
                 f"qubit_locations has {len(qubit_locations)} element(s) but the "
                 f"observable(s) have {cog.general_observable.num_qubits} qubit(s)."
             )
+
+    # Find observable_measurements register
+    for reg in qc.cregs:
+        if reg.name == "observable_measurements":
+            obs_creg = reg
+            break
+    else:
```
caleb-johnson, line 314: Neat code here where the else sits outside the looped if ... I'll have to glance at how this works :)

circuit_knitting/cutting/cutting_experiments.py
```diff
@@ -273,19 +305,29 @@ def _append_measurement_circuit(
...
+        raise ValueError(
+            '"observable_measurements" register is the wrong size '
+            "for the given commuting observable group "
+            f"({obs_creg.size} != {len(pauli_indices)})"
+        )
+
     if not inplace:
         qc = qc.copy()
 
-    # If the circuit has no measurements, the Sampler will fail.  So, we
```
caleb-johnson, line 279: Could you quickly explain why we can remove this? I may figure it out as I continue the review, but just in case ;)

circuit_knitting/cutting/cutting_reconstruction.py
```diff
@@ -111,21 +109,7 @@ def reconstruct_expectation_values(
             for k, cog in enumerate(so.groups):
                 quasi_probs = results_dict[label].quasi_dists[i * len(so.groups) + k]
                 for outcome, quasi_prob in quasi_probs.items():
-                    try:
```
caleb-johnson, line 114: niiiiice

caleb-johnson (commented):
Dropping a couple thoughts here. Will finish this up in a bit. Looking great!

circuit_knitting/cutting/cutting_experiments.py
garrison: Got moved to a new function `_get_pauli_indices`, because it needs checked now in two places. :)

caleb-johnson (approved):
Great idea, thanks for puttin this PR together. Makes thing much more straightforward :)

garrison:
@Ibrahim-Shehzad: since @caleb-johnson already approved everything before f4dec604669fbe3d1103256221cb6d10ea1b32c7, and f4dec604669fbe3d1103256221cb6d10ea1b32c7 primarily updates a test that you wrote, I think it makes most sense for you to review the changes in f4dec604669fbe3d1103256221cb6d10ea1b32c7 and approve this PR if those changes look good to you.  Thanks!

test/cutting/test_cutting_roundtrip.py
```diff
@@ -179,28 +179,56 @@ def test_cutting_exact_reconstruction(example_circuit):
...
 
+    This is related to https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/issues/422.
     """
+    # Create a circuit to cut
+    qc = QuantumCircuit(3)
+    append_random_unitary(qc, [0, 1])
+    append_random_unitary(qc, [2])
+    qc.rxx(np.pi / 3, 1, 2)
+    append_random_unitary(qc, [0, 1])
+    append_random_unitary(qc, [2])
```
Ibrahim-Shehzad, line 193: What motivates the choice of this particular circuit here?

garrison: It's sufficiently simple, but not trivial in any way, either.  It has two qubits involved in cutting, and a third qubit that gets entangled both before and after the cut but is otherwise "along for the ride."

garrison: It's also a special case of a circuit generated by the `example_circuit` fixture, which was used before.  But we don't need to spend time testing against every single one of them.  Just one should be sufficient, and is enough to fail without me fixing the line with `_get_pauli_indices`.

Ibrahim-Shehzad: So is there something significant that is lost if we do away with appending random unitaries before and after adding the rxx?

test/cutting/test_cutting_roundtrip.py
```diff
@@ -179,28 +179,56 @@ def test_cutting_exact_reconstruction(example_circuit):
...
         label: sampler.run(subexperiments[label]).result()
         for label, sampler in samplers.items()
     }
-    _ = results
+    reconstructed_expvals = reconstruct_expectation_values(
+        results, coefficients, subobservables
+    )
+
+    if is_exact_sampler:
+        # Determine exact expectation values
```
Ibrahim-Shehzad, line 214: Apologies for this naive question but why use `ExactSampler()` and not just `Sampler()` here?

test/cutting/test_cutting_roundtrip.py
garrison: If we got rid of the random unitaries before the cut gate, we'd only be testing that cutting works under the assumption that |0> is the input state, so there could be bugs that fall through.

Likewise, if we got rid of the random unitaries after the cut gate, then we'd only be ensuring that the post-cut state has the right amplitude in the Z basis, again, a much more limited test compared to the current, which should only reliably work if the post-cut state is correct in all bases.

test/cutting/test_cutting_roundtrip.py
garrison: As a result of the `pytest.mark.parametrize` decorator to this function, this test gets run twice, once with `Sampler()` and once with `ExactSampler()`.  The run using `Sampler()` has `is_exact_sampler is False`, so it's really just a smoke test: it makes sure the test runs without error, but it doesn't really test the result.  I made this choice because the Sampler is going to have shot noise, so the result it provides is only going to be approximate.  So, there is a second run with `ExactSampler()` with `is_exact_sampler is True`.  This "exact sampler" will return the exact probability distribution, without shot noise.  So we can inspect the reconstructed results and expect them to be exactly correct, up to roundoff error.

Ibrahim-Shehzad: I see, the part about shot noise addresses my question, thanks for the clarification.

test/cutting/test_cutting_roundtrip.py
Ibrahim-Shehzad: Got it, thank you so patiently clarifying this.

Ibrahim-Shehzad (approved):
This looks good to me!

PR #408, opened 2023-09-08
Unwrap `Tensor` objects from Qiskit Nature as necessary

garrison:
This doesn't quite pass CI yet.  At least one more change is needed somewhere.

Fixes #406.

PR #435, opened 2023-10-09
Temporarily pin qiskit-nature < 0.7

garrison:
Until #408 is merged (and #406 is thus fixed), we know that CKT will not work with Qiskit Nature 0.7 (a release that is expected some time soon).  We might as well preemptively mark our CKT releases until #408 is merged as being incompatible with Qiskit Nature 0.7.  Otherwise, users might be unpleasantly surprised once Qiskit Nature 0.7 is released and entanglement forging no longer works.

As an alternative, we _could_ just prioritize getting #408 in shape to merge, but there's no harm in having this pinned until that happens.

TODO
- [x] add a release note

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ dependencies = [
     "rustworkx>=0.12.0",
     "qiskit-aer>=0.12.0",
     "qiskit>=0.43.0",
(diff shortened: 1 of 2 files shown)


caleb-johnson (approved):
Thanks for taking care of this!

PR #83, opened 2023-04-05
Migrate entanglement forging to Qiskit Nature 0.5.0

caleb-johnson:
Fixes #89 
Fixes #94 

This PR removes the deprecated Qiskit Nature API calls from the entanglement forging source code.

This PR removes support for deprecated Qiskit Nature classes as inputs. See the release note for more details.

Considerations/Stray Thoughts:
- The EF with quantum serverless notebook tutorial is breaking on latest/dev branches in CI due to an incompatibility between the new `ElectronicStructureProblem` class in Nature 0.6.0 and quantum serverless. The serverless folks probably need to add support for a new class or two to support serialization of the new classes coming from Nature.  I've made an issue [here](https://github.com/Qiskit-Extensions/quantum-serverless/issues/413). I've turned off testing of this notebook for now. I think that's the best we can do until a QServerless fix is available.

Tasks before merging:
- [x] Make issue for adding the serverless tutorial back to the development workflow in CI when quantum serverless fix is available.

- [x] Ensure MO coeff handling is done in the most reasonable way. Work with Nature devs to review.

- [x] Update rst files in docs with new Nature API calls
- [x] merge in latest `main` and revert #120 on this branch

PR #449, opened 2023-11-06
Update forging how-to to Qiskit Nature 0.5

garrison:
This was overlooked in #83.  It is now causing CI errors in the docs workflow (see #448), now that a Qiskit version without `PauliTable` has been released.

### DIFF
--- a/docs/entanglement_forging/how-tos/specify-problem.rst
+++ b/docs/entanglement_forging/how-tos/specify-problem.rst
@@ -2,30 +2,31 @@
 How to specify the problem
 ##########################
 
-To specify the problem, we follow the Qiskit Nature workflow. We set up the ``molecule`` object, specify the driver, and instantiate the ``ElectronicStructureProblem``.
+To specify the problem, we follow the Qiskit Nature workflow. We set up the ``molecule`` object, specify the driver, and instantiate the :class:`.ElectronicStructureProblem`.
 
 We first require the following modules:
 
 .. jupyter-execute::
 
-   from qiskit_nature.drivers import Molecule
-   from qiskit_nature.drivers.second_quantization import PySCFDriver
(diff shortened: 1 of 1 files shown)


docs/entanglement_forging/how-tos/specify-problem.rst
```diff
@@ -2,39 +2,42 @@
...
 
    driver = PySCFDriver.from_molecule(molecule=molecule, basis="sto6g")
 
 Here, the driver is an algorithm class that knows how to calculate the second quantized operators.
 
-Finally, we instantiate the ``ElectronicStructureProblem``, a class which wraps a number of different types of drivers:
+Finally, we instantiate the :class:`.ElectronicStructureProblem`, a class which wraps a number of different types of drivers:
 
 .. jupyter-execute::
+   :stderr:
```
garrison, line 40: This cell really, really wants to spit something to stderr, and the only way I could figure out not to have it cause the sphinx build to fail was to put this line here.

caleb-johnson: Interesting, what is it printing out?

garrison: Here's the warning:

> UserWarning: Since PySCF-2.3, B3LYP (and B3P86) are changed to the VWN-RPA variant, corresponding to the original definition by Stephens et al. (issue 1480) and the same as the B3LYP functional in Gaussian. To restore the VWN5 definition, you can put the setting "B3LYP_WITH_VWN5 = True" in pyscf_conf.py

This is also showing up in some Qiskit Nature notebooks, such as at https://qiskit.org/ecosystem/nature/migration/0.5_c_electronic_structure.html.

caleb-johnson (approved):
Looks good, thank you!

PR #458, opened 2023-11-21
Remove unnecessary resets in `generate_cutting_experiments`

### LINKED ISSUE #452: Cut `Move` instructions do not work on hardware backends unless they support the `Reset` operation
Currently, performing wire cutting by using cut versions of the `Move` operation works only on backends that support the `Reset` instruction.  This includes `ibmq_qasm_simulator`, but some hardware backends do not currently support this operation.

Here is my proposed plan for dealing with this.  Note that, by following these steps, we will be able to support cut `Move` operations on all backends if qubits are not re-used.  However, one must use a backend that supports `Reset` any time one wishes to re-use qubits.

- [x] Create and document the use of a transpiler pass that will elide unnecessary `Reset`s.  Any number of resets that appear at the beginning of a qubit line can be elided, as the qubit is already in the |0> state.  Additionally, any number of resets that appear at the end of a qubit line can be elided as well.  Together, this means we can elide all `Reset` operations necessary for the `Move` subexperiments, as long as we don't re-use any of the qubits in a subsequent `Move` instruction.  (The current tutorial _does_ re-use qubits, however, which brings me to the following point...) - done, see #458.
- [ ] Swap the content of the wire cutting tutorial and how-to.  @caleb-johnson has suggested this before, but I feel that the current issue is overwhelming evidence that we should teach people to use `CutWire` in the tutorial, and to instead have a how-to guide if people want to cut `Move` operations directly (and potentially re-use qubits on hardware that supports `Reset`).
- [ ] Document clearly that re-using qubits for the cut `Move` instructions is only supported by backends that support the `Reset` operation.  (Or, potentially, the `Reset` operation could be implemented on any backend that supports dynamic circuits, if we were to add support for this use case.)

garrison:
Fixes #452.

I'd still like to swap the tutorial and how-to as suggested in the second point at https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/issues/452#issue-1991864801, but this can be done independently in a follow-up PR.

### DIFF
--- a/circuit_knitting/cutting/cutting_experiments.py
+++ b/circuit_knitting/cutting/cutting_experiments.py
@@ -90,7 +90,6 @@ def generate_cutting_experiments(
             appended to the gate label so they may be associated with other gates belonging
             to the same cut.
         ValueError: :class:`SingleQubitQPDGate` instances are not allowed in unseparated circuits.
-
     """
     if isinstance(circuits, QuantumCircuit) and not isinstance(observables, PauliList):
         raise ValueError(
--- a/circuit_knitting/cutting/cutting_experiments.py
+++ b/circuit_knitting/cutting/cutting_experiments.py
@@ -19,8 +19,11 @@ from collections.abc import Sequence, Hashable
 import numpy as np
 from qiskit.circuit import QuantumCircuit, ClassicalRegister
 from qiskit.quantum_info import PauliList
+from qiskit.transpiler import PassManager
+from qiskit.transpiler.passes import RemoveResetInZeroState, DAGFixedPoint
 
 from ..utils.iteration import strict_zip
+from ..utils.transpiler_passes import RemoveFinalReset, ConsolidateResets
 from ..utils.observable_grouping import ObservableCollection, CommutingObservableGroup
 from .qpd import (
     WeightType,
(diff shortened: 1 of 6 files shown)


circuit_knitting/cutting/cutting_experiments.py
```diff
@@ -156,6 +165,25 @@ def generate_cutting_experiments(
...
+    # (which can arise when re-using qubits) into a single reset.
+    pass_manager = PassManager()
+    pass_manager.append(
+        [
+            RemoveResetInZeroState(),
+            RemoveFinalReset(),
+            ConsolidateResets(),
+            DAGFixedPoint(),
+        ],
+        do_while=lambda property_set: not property_set["dag_fixed_point"],
```
caleb-johnson, line 182: So you repeat these passes until this condition is False. What exactly triggers this? I haven't come up to speed on how control flow operations work yet.

circuit_knitting/utils/__init__.py
```diff
@@ -59,6 +59,12 @@
 =============================================================
 
 .. automodule:: circuit_knitting.utils.transforms
+
+===================================================================
+Transpiler passes (:mod:`circuit_knitting.utils.transpiler_passes`)
+===================================================================
+
+.. automodule:: circuit_knitting.utils.transpiler_passes
```
caleb-johnson, line 67: 👀 fancy

caleb-johnson (approved):
Cool! Thanks for setting this up. LGTM I'm sure we will continue to have more ideas for transpiler passes

I just had one question about the control flow stuff

circuit_knitting/cutting/cutting_experiments.py
garrison: The `DAGFixedPoint` pass [sets this property](https://qiskit.org/documentation/stubs/qiskit.transpiler.passes.DAGFixedPoint.html).  It's neat; I learned about it only by reading [the tests](https://github.com/Qiskit/qiskit/blob/e91947e1b55f6e661c683fba72ab66c5f92d3aec/test/python/transpiler/test_remove_reset_in_zero_state.py#L81-L102) for the `RemoveResetInZeroState` pass in qiskit.

PR #463, opened 2023-12-07
Update `UnitaryGate` import now that `qiskit.extensions` is deprecated

garrison:
`qiskit.extensions` is deprecated starting with Qiskit 0.45, and the `UnitaryGate` class has moved to `qiskit.circuit.library`. However, this class did not exist at this location prior to Qiskit 0.45, so this means having to bump the minimum Qiskit version as well.


garrison:
If we merge this before the next release, it will allow us to knock out a few `DeprecationWarning`s at the cost of requiring a more recent qiskit (i.e., one released in November rather than August).  I don't think this is a huge down side, but there's also no urgency to merge this before the next release.  (We'll definitely have to merge it before CKT can work with Qiskit 1.0, however.)

Issue #335, opened 2023-07-21
Queue all batches simultaneously in cutting


caleb-johnson:
execute_experiments should queue all of its batched jobs simultaneously. Currently, if a user wants to run two partitions on two different backends, they have to wait for one backend to finish before even queueing for the other backend.

This wasn't a problem before, as we were parallelizing; however, parallelizing the execution of the quantum experiments was causing hanging jobs fairly consistently.

We likely need to refactor the running of the batched experiments to kick off their jobs together, before waiting for them to each finish. Maybe break `_run_experiments_batch` into two functions.

PR #469, opened 2024-01-04
Update description of cutting tut 1

caleb-johnson:
The Qiskit Runtime doesn't behave well when you use primitives in parallel. The CKT tutorials which ran the circuit batches in parallel would randomly, and rather frequently, hang indefinitely. We never tracked down the cause, but we suspected the primitives were just not thread safe.

We removed the parallelization from the tutorials, but we did not update the description of tutorial 1 to reflect that. This PR updates that description.


Ibrahim-Shehzad:
LGTM!

Ibrahim-Shehzad (approved):
Looks good, thanks!

caleb-johnson:
I tried to submit all the jobs in parallel before retrieving each of their results in #335 , but I ran into CICD hanging there as well.

PR #471, opened 2024-01-09
Lo circuit cuts optimizer

caleb-johnson:
Create an interface for users to automatically find LO gate + wire cuts, given some cut criteria. This interface should be part of and compatible with `circuit_knitting.cutting`.


caleb-johnson:
Known issues:

- Problem with `cz`

```
from qiskit import QuantumCircuit
from circuit_knitting.cutting import find_cuts

circuit = QuantumCircuit(2)
circuit.cz(0, 1)


cut_circuit = find_cuts(circuit, {"rand_seed": 111}, {"qubits_per_QPU": 1, "num_QPUs": 2})
```

```
AttributeError: 'NoneType' object has no attribute 'upperBoundGamma'
```

circuit_knitting/cutting/cut_finding/circuit_interface.py
Ibrahim-Shehzad, line 1: The type hints added in this file are not fully complete. For example, in a few instances, I specify the type as `list` and not `list[...]`, I will have to return to this later.

circuit_knitting/cutting/cut_finding/search_space_generator.py
Ibrahim-Shehzad, line 1: A full type hint for `action_dict` would be `action_dict: dict[str, DisjointSearchAction]` but this would require importing `DisjointSearchAction` from `cutting_actions.py` which in turn relies on this module `(search_space_generator.py)` to be initialized. As a result, adding this type hint leads to a circular import. I will return to this issue later.

garrison: One possible solution is given at https://www.reddit.com/r/learnpython/comments/zzgkxj/comment/j2bmhlj/.

caleb-johnson: > One possible solution is given at https://www.reddit.com/r/learnpython/comments/zzgkxj/comment/j2bmhlj/.

Would you go directly for this solution, or do you consider this error an indication the design should be modified?

circuit_knitting/cutting/cut_finding/search_space_generator.py
```diff
@@ -0,0 +1,208 @@
...
+    mincost_bound_func (lambda *args) can either be None or a function that
+    returns a cost bound that is compared to the minimum cost across all
+    vertices in a search frontier.  If the minimum cost exceeds the min-cost
+    bound, the search is terminated even if a goal state has not yet been found.
+    Returning None is equivalent to returning an infinite min-cost bound (i.e.,
+    min-cost checking is effectively not performed).  A mincost_bound_func that
+    is None is likewise equivalent to an infinite min-cost bound.
+    """
+
+    cost_func:  Callable  = None,
```
caleb-johnson, line 182: We can make an attempt to fill out these Callable types. For example, ``Callable[[ArgAType, ArgBType], ReturnType]``

Ibrahim-Shehzad: Yes, it's on my list of things to do.

circuit_knitting/cutting/cut_finding/circuit_interface.py
Ibrahim-Shehzad, line 1: Most of tests were based on the older version of `CircuitInterface` which did not use the `CircuitElement` `NamedTuple`. The new version causes those tests to fail. The test inputs will need to be modified. Since `CircuitElement` now has fields which correspond to `params` and `gamma`, I am wondering what the best way would be to specify single qubit gates, like `h` and `x` etc.

test/cutting/cut_finding/test_circuit_interfaces.py
Ibrahim-Shehzad, line 1: @caleb-johnson I updated the test circuits that were being used in this test to make them consistent with the new interface which uses `CircuitElement`. To do this consistently, I assigned `params = []` to gates that do not depend on any parameters. Moreover, for single qubit gates, I set `gamma = 1`. To distinguish `barriers` placed across the whole circuit from `barriers` placed across a subset of the gates, I defined the latter as a `CircuitElement` object as well (with `params` giving the qubits the barrier is placed across and `gamma` being set to 1). The tests in this particular module now run fine. Let me know what your thoughts are on all of this.

circuit_knitting/cutting/cut_finding/search_space_generator.py
garrison: @caleb-johnson, I agree, there is almost certainly something that should be reconsidered here.

Ibrahim-Shehzad: I've gotten it working using the 'fix' linked above; a different design wasn't coming to mind immediately but we can discuss that more.

test/cutting/cut_finding/test_circuit_interfaces.py
Ibrahim-Shehzad: For single qubit gates, I decided to go with `gamma = None` instead. The most recent commit reflects that.

circuit_knitting/cutting/cut_finding/search_space_generator.py
Ibrahim-Shehzad: For this particular module, commit 551237d87113102c741ae1aa1d84a4c519472da9 now has these Callable types filled in.

test/cutting/cut_finding/test_disjoint_subcircuits_state.py
```diff
@@ -227,8 +226,10 @@ def test_CutBothWires(testCircuit):
 
     assert next_state.getSearchLevel() == 1
 
-    assert next_state.lowerBoundGamma() == 9  # 3^n scaling.
+    assert (
+        next_state.lowerBoundGamma() == 9
```
Ibrahim-Shehzad, line 230: @caleb-johnson I have kept around the bell-pair functionality for wire cuts which is what gives the 3^n scaling described here in `lowerBoundGamma()`. This was there for gate cuts as well but I noticed that you took it out recently. My feeling is that it is okay (good?) to keep this around, even though currently we do not use it to calculate our overall `gamma` since we only support LO cuts. What are your thoughts?

Ibrahim-Shehzad:
@caleb-johnson I added type annotations and (nominally) got rid of all the mypy errors. However, some of it was a little rushed and I am not entirely satisfied with it just yet. Please feel free to call out anything that you think can be improved right away. 

Going forward, I think what remains to be done is:

1. Add tests for `cut_finding.py`
2. Remove all the linting errors.
3. Improve the function and class names (in particular, get rid of the camel casing). 
4. Decide on public vs private methods and finalize doc strings.

I will start an issue listing these things as well.

circuit_knitting/cutting/cut_finding/cut_finding.py
```diff
@@ -0,0 +1,102 @@
...
+    optimizer = LOCutsOptimizer(interface, opt_settings, constraint_settings)
+
+    # Find cut locations
+    opt_out = optimizer.optimize()
+
+    wire_cut_actions = []
+    gate_ids = []
+
+    opt_out = cast(DisjointSubcircuitsState, opt_out)
+    opt_out.actions = cast(list, opt_out.actions)
```
caleb-johnson, line 77: What error were you seeing that caused you to have to cast `opt_out` here?

circuit_knitting/cutting/cut_finding/cut_optimization.py
```diff
@@ -0,0 +1,326 @@
...
+    # account any user-specified constraints that might have been
+    # placed on how the current entangling gate is to be handled
+    # in the search
+    gate = gate_spec[1]
+    gate = cast(CircuitElement, gate)
+    if len(gate.qubits) == 2:
+        action_list = func_args.search_actions.get_group("TwoQubitGates")
+    else:
+        raise ValueError(
+            "In the current version, only the cutting of two qubit gates is supported."
```
caleb-johnson, line 99: Currently, if I pass in a circuit with only 3-qubit CCXGates, this will throw an error because it will try to cut those gates. I think it'd be better if the optimizer just didn't consider these gates valid for cutting and ignored them. I'd rather throw a more general error if the optimizer can't find a valid cut scheme only from cutting 2-qubit gates and wires.

circuit_knitting/cutting/cut_finding/cut_finding.py
```diff
@@ -0,0 +1,102 @@
...
+    circ_out = cut_gates(circuit, gate_ids)[0]
+
+    # Insert all the wire cuts
+    counter = 0
+    for action in sorted(wire_cut_actions, key=lambda a: a[1][0]):
+        if action[0].get_name() == "CutTwoQubitGate":
+            continue
+        inst_id = action[1][0]
+        qubit_id = action[2][0][0] - 1
+        circ_out.data.insert(
```
caleb-johnson, line 96: I currently make no distinction between "left wire cut" and "right wire cut", which means I just place the cut wire on the index of the gate the wire cut references, which means it is always essentially a "left wire cut". I'm fairly sure this is an error. Noting here to follow up

caleb-johnson: I'm less certain this is a bug after thinking about it, considering that I haven't encountered any problems separating the circuit under this logic. Any thoughts, @Ibrahim-Shehzad ?

circuit_knitting/cutting/cut_finding/cut_finding.py
Ibrahim-Shehzad: I think there was a mypy error here, because of the fact that `opt_out` is also allowed to be `None`.

circuit_knitting/cutting/cut_finding/cut_finding.py
Ibrahim-Shehzad: Perhaps I misunderstood your question but, in general, I don't think we can get away without specifying whether the first input wire to the gate is cut ("left wire cut") or the second input wire ("right wire cut"). I do think that it makes a difference. Was this what you were getting at?

circuit_knitting/cutting/cut_finding/cut_optimization.py
Ibrahim-Shehzad: > I think it'd be better if the optimizer just didn't consider these gates valid for cutting and ignored them

Agreed. I will get it working this way.

Ibrahim-Shehzad: @caleb-johnson , I just remembered that my thought process behind this was that if the user gave us a (contrived) circuit with, say, just one 3-qubit gate and asked us to cut it, it should through an error. Perhaps we should go with what you're advocating and say that only 2-qubit gates can be cut somewhere in the documentation?

circuit_knitting/cutting/cut_finding/cut_finding.py
caleb-johnson: I think the "Left" and "Right" are confusing me because they aren't referencing which side of an operation to cut, they are referencing which qubit to cut with respect to some instruction. You always cut before the instruction, regardless of "Left" or "Right".

That seems to be what I implemented here, and it seems to always work, so I assume that is the logic. Does that sound right?

Ibrahim-Shehzad: > ...they aren't referencing which side of an operation to cut, they are referencing which qubit to cut with respect to some instruction. You always cut before (to the left of) the instruction, regardless of "Left" or "Right".

That is indeed correct.

circuit_knitting/cutting/cut_finding/cut_optimization.py
caleb-johnson: Yes, that's exactly the case I created to test this. I made a 7-qubit circuit with a ladder of CCXGates. I would expect the optimizer to either tell me it couldn't find a solution or to return me the solution, regardless of the input circuit. How hard would it be to make it behave like that?

Ibrahim-Shehzad: Do you mean how hard it would be to get it to cut arbitrary qubit gates? There was some machinery there for cutting multi-qubit gates at one point and we took it out (after some discussion) because we said that cutting 2 qubit gates would be the most immediate use case. Another motivation for doing this was that the transpiler converts all input gates into single and two qubit gates.

Ibrahim-Shehzad: Having looked through the changes I've made in other places in the code (since we discussed removing support for gates with 3 or more inputs), the only consistent thing to do would be to throw an error when a gate with more than 2 qubits is encountered in an input circuit. Changing this would involve adding back some the machinery I talked about above, even if the goal is just apply that gate and ignore cutting it. We can discuss whether than needs to be done but it would be a little non-trivial.

caleb-johnson: I think it's OK for the MVP to not accept inputs containing ``N>2``-qubit gates. I'll catch that error in `find_cuts` and tell users to decompose their `N`-qubit gates into single and two-qubit operations

test/cutting/cut_finding/test_best_first_search.py
```diff
@@ -0,0 +1,101 @@
+from __future__ import annotations
```
caleb-johnson, line 1: The tests look great, thank you for all this hard work.

I wonder why you didn't want to use the unittest framework we used everywhere else in the cutting tests for the tests in this package?

Ibrahim-Shehzad: I don't actually have a great reason for that :), perhaps the only thing I can say is that these tests were initially developed on my private cco repo and then ported over.

docs/circuit_cutting/tutorials/04_automatic_cut_finding.ipynb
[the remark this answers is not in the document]
garrison: > `optimization_settings = {"rand_seed": 111}`

Same with this one -- and it should be either `random_seed` or just `seed` (no need to introduce a special abbreviation here).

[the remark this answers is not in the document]
garrison: A comment about the name `num_QPUs`: in the strictly LO case, I believe we can execute the resulting circuits on a single QPU even if the `num_QPUs` provided to the optimizer is higher -- is that correct?  The idea is that things can be executed sequentially on the same processor.  I am trying to understand whether the `num_QPUs` knob should be thought of as controlling the number of hardware devices you have access to, or if it can instead be proportional to the maximum number of sequential subcircuit executions you are willing to wait for, even if you only have access to one QPU.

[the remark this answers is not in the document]
Ibrahim-Shehzad: So we're not allowing for sequential execution on the same processor (since there is no notion of qubit reuse yet). It is indeed the number of hardware devices you have access to. So for example, if you have a 5 qubit circuit and 2 qpu's with 3 and 2 qubits respectively, then the circuit will be cut into sub-circuits of width 3 and 2. If instead you have a 3 qubit circuit, then the optimal case would be to not cut at all and run the entire, un-cut circuit on the 3 qubit qpu. Note, however, that even in this second case, the optimizer does (internally) keep track of suboptimal solutions, one of which would be to cut it up as 2 + 1.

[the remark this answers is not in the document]
garrison: In both the LO and one-way LOCC cases, you can simulate additional QPUs by just running additional jobs sequentially on a single QPU.  This should be independent of whether qubit re-use is supported or not.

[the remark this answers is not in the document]
Ibrahim-Shehzad: I see, I think I meant that the cut finder (at present) doesn't support qubit reuse or sequential execution on the same processor. I should carefully recheck this though.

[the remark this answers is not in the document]
Ibrahim-Shehzad: > num_QPUs really just tells the optimizer max number of subcircuits to produce when cutting.

This is correct. I realize I misunderstood the thrust of your initial question, @garrison .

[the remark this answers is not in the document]
garrison: > num_QPUs really just tells the optimizer max number of subcircuits to produce when cutting.

In that case, I think we can find a better name than `num_QPUs`.  Perhaps something like `max_subcircuits`?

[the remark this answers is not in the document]
Ibrahim-Shehzad: `num_QPUs` sounded a little more evocative to me in this context. I think it was also picked with the eventual progression to LOCC in mind. I think I am open to going with your suggestion though.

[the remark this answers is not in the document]
Ibrahim-Shehzad: >It'd be nice if this device_constraints were its own class, rather than a dict with special keys.
@garrison , is the idea here that the user would pass in an object with the two fields `qubits_per_QPU` and `max_subcircuits` specified?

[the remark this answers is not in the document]
Ibrahim-Shehzad: I will go ahead and remove all references to it.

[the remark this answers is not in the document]
garrison: > @garrison , is the idea here that the user would pass in an object `DeviceConstraints` with the two fields `qubits_per_QPU` and `max_subcircuits` specified?

Yes (except, for the reasons you mention, with only the `qubits_per_QPU` field for the time being).

circuit_knitting/cutting/cutting_decomposition.py
```diff
@@ -260,3 +270,119 @@ def decompose_observables(
...
+    # Insert all the wire cuts
+    counter = 0
+    for action in sorted(wire_cut_actions, key=lambda a: a[1][0]):
+        inst_id = action.gate_spec.instruction_id
+        # args[0][0] will be either 1 (control) or 2 (target)
+        qubit_id = action.args[0][0] - 1
+        circ_out.data.insert(
+            inst_id + counter,
+            CircuitInstruction(
+                CutWire(), [circuit.data[inst_id].qubits[qubit_id]], []
```
garrison, line 363: Let's make a test case that would fail if this still said `inst_id + counter` rather than `inst_id` (fixed in 25a9d06ddc31d32ddc119ac36cdff74e8a48eb01).

circuit_knitting/cutting/cutting_decomposition.py
```diff
@@ -260,3 +270,119 @@ def decompose_observables(
...
+        )
+        counter += 1
+        if action.action.get_name() == "CutBothWires":
+            # There should be two wires specified in the action in this case
+            assert len(action.args) == 2
+            qubit_id2 = action.args[1][0] - 1
+            circ_out.data.insert(
+                inst_id + counter,
+                CircuitInstruction(
+                    CutWire(), [circuit.data[inst_id + counter].qubits[qubit_id2]], []
```
garrison, line 374: Given the above, are you absolutely sure the ` + counter` on this line is correct?

Ibrahim-Shehzad: I think that needs to be removed as well.

circuit_knitting/cutting/cutting_decomposition.py
```diff
@@ -294,9 +294,6 @@ def find_cuts(
             - seed: A seed for the pseudorandom number generator used by the optimizer
         constraints: Options for specifying the constraints for circuit cutting
 
-            - qubits_per_QPU: The maximum number of qubits each subcircuit can contain
```
caleb-johnson: WHy did we delete this again?

Ibrahim-Shehzad: Oops, that was a mistake on my part. Will fix this.

caleb-johnson:
When we squash these commits into a single commit, we need to make sure Ed is a co-author.

caleb-johnson:
Has @pednault released any papers externally that we should cite in the docs?

garrison:
> When we squash these commits into a single commit, we need to make sure @pednault is a co-author.

I expect him to be because 33a48fbb52d1583aba2ef145b11d80b529b36457 is in the history here.

Ibrahim-Shehzad:
>Has @pednault released any papers externally that we should cite in the docs?

I don't think so.

garrison:
> Has @pednault released any papers externally that we should cite in the docs?

From arXiv, it looks like Ed's most recent paper is on LOCC wire cutting.  He has posted nothing so far there about cut finding:  https://arxiv.org/search/?query=pednault%2C+e&searchtype=author

circuit_knitting/cutting/cutting_decomposition.py
```diff
@@ -260,3 +270,115 @@ def decompose_observables(
     }
 
     return subobservables_by_subsystem
+
+
+def find_cuts(
```
garrison, line 275: I'd prefer to see this function introduced in a new/different file, since it's a bit independent of the existing code in this file.  (Actually, it depends on the existing code in this module through `cut_gates`, and moving to a separate file will help us make sure the dependency remains in a one-way direction, rather than have a risk of becoming spaghetti over time.)

caleb-johnson: It was originally in `cut_finding` package, and I decided to move it up alongside the other modules in our cutting workflow. I chose `cutting_decomposition` because it felt like the most natural module. I've created a new module, `find_cuts.py` that includes this function now.

circuit_knitting/cutting/find_cuts.py
```diff
@@ -0,0 +1,143 @@
...
+    """Find cut locations in a circuit, given optimization settings and QPU constraints.
+
+    Args:
+        circuit: The circuit to cut. The circuit must contain only single two-qubit
+            gates.
+        optimization: Options for controlling optimizer behavior. Currently, the optimal
+            cuts are arrived at using Dijkstra's best-first search algorithm. The specified
+            parameters are:
+
+            - max_gamma: Specifies a constraint on the maximum value of gamma that a
```
caleb-johnson, line 47: We probably don't need to explicitly state all this in the docstring anymore since we are exposing the settings classes in the docs

Ibrahim-Shehzad:
Thoughts now on merging this PR, @garrison , @caleb-johnson ?

caleb-johnson:
> Thoughts now on merging this PR, @garrison , @caleb-johnson ?

I'll wait for you both to give the 👍 , and I'll give it one more pass before merging

Ibrahim-Shehzad:
I think we're good to go, from my perspective.

circuit_knitting/cutting/cutting_decomposition.py
Ibrahim-Shehzad: 'tis done, [f91fac0](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/471/commits/f91fac09f6bceef223943d53d68e3d3f60e24b70).

circuit_knitting/cutting/cutting_decomposition.py
```diff
@@ -13,9 +13,9 @@
 
 from __future__ import annotations
 
+from typing import NamedTuple
 from collections import defaultdict
 from collections.abc import Sequence, Hashable
-from typing import NamedTuple
```
garrison, line 18: It'd be nice to revert this change, which would mean one fewer file touched.  But not a big deal.  I am splitting hairs.

circuit_knitting/cutting/cut_finding/cco_utils.py
```diff
@@ -0,0 +1,161 @@
...
+    )
+
+    if best[-1] is not None:
+        return greedy_best_first_search(best[-1], search_space_funcs, *args)
+
+    # The else block below covers a rare edge case
+    # which needs a clever circuit to get tested.
+    # Excluding from test coverage for now.
+    else:  # pragma: no cover
+        return None
```
garrison, line 161: Can anything else be said about this "rare edge case"?

circuit_knitting/cutting/cut_finding/cco_utils.py
```diff
@@ -0,0 +1,161 @@
...
+
+    Args:
+    circuit: an instance of :class:`qiskit.QuantumCircuit` .
+
+    Returns:
+    circuit_list_rep: list of circuit instructions represented in a form that is compatible with
+    :class:`SimpleGateList` and can therefore be ingested by the cut finder.
+
+    TODO: Extend this function to allow for circuits with (mid-circuit or other)
+    measurements, as needed.
```
garrison, line 45: I'm not suggesting we do this now, but it doesn't sound like it will be complicated to handle this case when the need arises. (Right?)  In particular, we will need this some day if the cut finder is going to support cutting a circuit in order to reconstruct a probability distribution (i.e., once we have an updated version #428 that has been brought up to date with `main`).

Ibrahim-Shehzad: I'd have to see how mid-circuit measurements enter circuit instructions but I do imagine that it shouldn't be complicated to handle. I also just noticed a typo in this doc string, will fix and push.

circuit_knitting/cutting/cut_finding/cco_utils.py
Ibrahim-Shehzad: Mostly just that I wasn't able to find any such cases :)

test/cutting/test_find_cuts.py
```diff
@@ -0,0 +1,72 @@
...
+            constraints = DeviceConstraints(qubits_per_subcircuit=4)
+
+            _, metadata = find_cuts(
+                circuit, optimization=optimization, constraints=constraints
+            )
+            cut_types = {cut[0] for cut in metadata["cuts"]}
+
+            assert len(metadata["cuts"]) == 2
+            assert {"Wire Cut", "Gate Cut"} == cut_types
+            assert np.isclose(127.06026169, metadata["sampling_overhead"], atol=1e-8)
```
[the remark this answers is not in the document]
garrison: From looking at the random circuit code, it appears that adding a single gate to the standard library of gates used for constructing the circuits would mean a different result:

https://github.com/Qiskit/qiskit/blob/f81e5a87b298a79cc0e52997e3d4e4b7fbfce713/qiskit/circuit/random/utils.py#L153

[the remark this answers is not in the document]
garrison: I checked with Matthew Treinish: "I'd say with in a single minor release series you can rely on the output being exactly the same, but between minor or major versions I don't think we can guarantee that."

So, in other words, we run the risk that this test might begin failing when Qiskit 1.1 or some later non-patchlevel release comes out.

I don't see a reason this absolutely needs to hold up this PR, but it's something that we should fix at some point.

As ugly as it would be, specifying the equivalent circuit explicitly in code would be preferable.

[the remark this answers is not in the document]
Ibrahim-Shehzad: Think this may even be applicable to [this](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/LO-circuit-cuts-optimizer/docs/circuit_cutting/tutorials/04_automatic_cut_finding.ipynb) tutorial.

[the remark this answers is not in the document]
garrison: I feel better about the tutorial because in principle the code should still work even if the circuit changes (it may just give a different result).  With the test, it means a CI or test failure if somebody uses a version of Qiskit that made the change, because the tests rely explicitly on properties of that circuit.

Then again, perhaps you are saying that with a slightly different circuit, it might be that no suitable cuts could be found?  (Or perhaps even way too many, after searching for a long time?)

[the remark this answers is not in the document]
Ibrahim-Shehzad: I actually agree that with the tutorial it's going to be okay because I don't expect crazy hard-to-cut circuits popping up when the number of qubits is that small. I just flagged it as something to potentially consider but I agree that we can keep the tutorial as is.

circuit_knitting/cutting/cutting_decomposition.py
```diff
@@ -13,6 +13,7 @@
 
 from __future__ import annotations
 
+
```
garrison, line 16:
```suggestion
```
will fully remove this file from the diff

circuit_knitting/cutting/cut_finding/cco_utils.py
```diff
@@ -0,0 +1,161 @@
...
+    )
+
+    if best[-1] is not None:
+        return greedy_best_first_search(best[-1], search_space_funcs, *args)
+
+    # The else block below covers a rare edge case
+    # which needs a clever circuit to get tested.
+    # Excluding from test coverage for now.
+    else:  # pragma: no cover
+        return None
```
garrison, line 161: I clarified the comment and suggest using a guarded-if pattern, following e.g. https://medium.com/lemon-code/guard-clauses-3bc0cd96a2d3, to make the normal case go straight through and have the exceptional case covered by the `if` statement.

```suggestion
    if best[-1] is None:  # pragma: no cover
        # This covers a rare edge case.
        # We have so far found no circuit that triggers it.
        # Excluding from test coverage for now.
        return None

    return greedy_best_first_search(best[-1], search_space_funcs, *args)
```

circuit_knitting/cutting/cut_finding/cco_utils.py
garrison: Conversation continued at https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/471#discussion_r1542965288.

releasenotes/notes/automatic-cut-finding-696556915e347138.yaml
```diff
@@ -0,0 +1,18 @@
+---
+features:
+  - |
+    Added a cut-finder function, :func:`circuit_knitting.cutting.find_cuts`, for automatically
+    identifying locations to place LO gate and wire cuts such that the circuit is
+    separable and runnable, given the maximum number of qubits per subcircuit.
```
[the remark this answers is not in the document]
Ibrahim-Shehzad: Have changed this locally, will commit and push soon.

circuit_knitting/cutting/cut_finding/cco_utils.py
Ibrahim-Shehzad: This is a good suggestion, thank you.

Ibrahim-Shehzad: 'Tis done :  [f64ca8a](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/471/commits/f64ca8a37aae7f645f2af31031b0b611e1921f3a)

test/cutting/test_find_cuts.py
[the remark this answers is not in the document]
Ibrahim-Shehzad: 'Tis done: [f64ca8a](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/471/commits/f64ca8a37aae7f645f2af31031b0b611e1921f3a)

releasenotes/notes/automatic-cut-finding-696556915e347138.yaml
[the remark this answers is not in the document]
Ibrahim-Shehzad: 'Tis done: [f64ca8a](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/471/commits/f64ca8a37aae7f645f2af31031b0b611e1921f3a)

circuit_knitting/cutting/find_cuts.py
```diff
@@ -0,0 +1,133 @@
...
+            counter += 1
+
+    # Return metadata describing the cut scheme
+    metadata: dict[str, Any] = {"cuts": []}
+    for i, inst in enumerate(circ_out.data):
+        if inst.operation.name == "qpd_2q":
+            metadata["cuts"].append(("Gate Cut", i))
+        elif inst.operation.name == "cut_wire":
+            metadata["cuts"].append(("Wire Cut", i))
+    metadata["sampling_overhead"] = opt_out.upper_bound_gamma() ** 2
```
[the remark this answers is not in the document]
Ibrahim-Shehzad: Let's maybe save this for a future PR?

circuit_knitting/cutting/cutting_decomposition.py
Ibrahim-Shehzad: Should be done now...

circuit_knitting/cutting/find_cuts.py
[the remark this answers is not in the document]
Ibrahim-Shehzad: Sounds good, done.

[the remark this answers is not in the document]
garrison: The follow-up issue is at https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/issues/516.

circuit_knitting/cutting/cutting_decomposition.py
garrison: I still see it in the overall diff at https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/471/files#diff-fcbd1b39891440c3ebf98536ee92ccbf9f7d84ed89db13422f95362ae8dbe3cb but not a big deal.

test/cutting/test_find_cuts.py
```diff
@@ -0,0 +1,80 @@
...
+)
+
+
+class TestCuttingDecomposition(unittest.TestCase):
+    def test_find_cuts(self):
+        with self.subTest("simple circuit"):
+            path_to_circuit = os.path.join(
+                os.path.dirname(__file__),
+                "..",
+                "qasm_circuits/circuit_find_cuts_test.qasm",
```
garrison, line 34: Your tests pass so it's fine, but fyi this is technically preferable so that it will work on a hypothesized platform that uses a path separator other than `/`.  (Historically, this would have included Windows, which uses `\`, but CI passes on Windows, so you're good in that specific case and all that we actually care about.)

```suggestion
                "qasm_circuits",
                "circuit_find_cuts_test.qasm",
```

Ibrahim-Shehzad: Oh yeah that makes more sense. I don't even know why I left it like that. Will change this.

tox.ini
```diff
@@ -27,6 +27,7 @@ commands =
   ruff check circuit_knitting/ docs/ test/ tools/
   nbqa ruff docs/
   autoflake --check --quiet --recursive circuit_knitting/ docs/ test/ tools/
+  black --diff circuit_knitting/ docs/ test/ tools/
```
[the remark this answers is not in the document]
Ibrahim-Shehzad: 'Tis done: [68b9c1c](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/471/commits/68b9c1cea37e54bb9f92ae0e85f02a927a111ef9)

test/cutting/test_find_cuts.py
Ibrahim-Shehzad: 'Tis done: [68b9c1c](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/471/commits/68b9c1cea37e54bb9f92ae0e85f02a927a111ef9)

caleb-johnson:
Closing in favor of #520

PR #475, opened 2024-02-07
Qiskit1.0

caleb-johnson:


pyproject.toml
```diff
@@ -30,9 +30,9 @@ dependencies = [
     "scipy>=1.5.2",
     "rustworkx>=0.13.0",
     "qiskit-aer>=0.12.0",
-    "qiskit>=0.44.1",
+    "qiskit==1.0.0b1",
```
mtreinish, line 33: I would use `1.0.0rc1` here  if you're testing for compatibility with 1.0.0. The `1.0.0b1`  release was just a very early snapshot for the quantum summit and the final api is very different in `rc1`

caleb-johnson: Thank you, I am seeing the more updated things now :)

I think `qiskit>=0.46<2.0` is the right thing to do to get us through the release?

.ruff.toml
```diff
@@ -1,9 +1,9 @@
 target-version = "py38"
-ignore = [
+lint.ignore = [
```
caleb-johnson, line 2: New ruff version is raising warnings and wants us to use this `lint` package

pyproject.toml
```diff
@@ -30,9 +30,9 @@ dependencies = [
     "scipy>=1.5.2",
     "rustworkx>=0.13.0",
     "qiskit-aer>=0.12.0",
-    "qiskit>=0.45.0, <1.0",
+    "qiskit>=0.46, <2",
```
caleb-johnson, line 33: I think this is the best way to specify the version to get us through the 1.0 release. I guess when 1.0 is released we can update the minimum here

pyproject.toml
```diff
@@ -91,7 +91,6 @@ docs = [
 ]
 notebook-dependencies = [
     "circuit-knitting-toolbox[cplex,pyscf]",
-    "quantum-serverless>=0.0.7",
```
caleb-johnson, line 94: I'd like to go ahead and remove both serverless tutorials and all dependency on that package. As Middleware definitions evolve, I see less of a need to feature serverless in the tutorials in this package

tox.ini
```diff
@@ -39,7 +39,7 @@ extras =
   nbtest
   notebook-dependencies
 commands =
-  pytest --nbmake --nbmake-timeout=3000 {posargs} docs/ --ignore=docs/_build --ignore=docs/entanglement_forging/tutorials/tutorial_2_forging_with_quantum_serverless.ipynb
+  pytest --nbmake --nbmake-timeout=3000 {posargs} docs/ --ignore=docs/_build --ignore=docs/entanglement_forging/tutorials/tutorial_1_getting_started.ipynb
```
caleb-johnson, line 42: We need to ignore this tutorial for now. See PR description for more info.

pyproject.toml
garrison: > I think this is the best way to specify the version to get us through the 1.0 release.

I agree.

circuit_knitting/cutting/cutting_experiments.py
```diff
@@ -165,25 +162,6 @@ def generate_cutting_experiments(
                 _append_measurement_circuit(new_qc, cog, inplace=True)
                 subexperiments_dict[label].append(new_qc)
 
-    # Remove initial and final resets from the subexperiments.  This will
```
caleb-johnson, line 168: We may need to put this back in and replace the `do_while`. Leaving it removed for now to focus on other compatibility issues.

pyproject.toml
caleb-johnson: 0.46 does not have package paths we need to fully update to 1.x (e.g. `qiskit.synthesis`), so I think we need to set the qiskit version here to the release candidate.

caleb-johnson: disregard above, I think that's wrong. Need to double check now that fewer things are broken

caleb-johnson:
@garrison , they have removed `do_while` from the `PassManager.append` method. Could you point me in the right direction of how we might recreate this functionality in the [one spot it is used in our code base + the test](https://github.com/search?q=repo%3AQiskit-Extensions%2Fcircuit-knitting-toolbox%20do_while&type=code)?

caleb-johnson:
> @garrison , they have removed `do_while` from the `PassManager.append` method. Could you point me in the right direction of how we might recreate this functionality in the [one spot it is used in our code base + the test](https://github.com/search?q=repo%3AQiskit-Extensions%2Fcircuit-knitting-toolbox%20do_while&type=code)?

From [qiskit docs](https://docs.quantum.ibm.com/api/qiskit/qiskit.transpiler.PassManager):

```
Creating flow controllers with flow_controller_conditions keyword arguments was deprecated. Instead, you must explicitly instantiate a controller and set the controller to passes argument.
```

garrison:
> @garrison , they have removed `do_while` from the `PassManager.append` method. Could you point me in the right direction of how we might recreate this functionality in the [one spot it is used in our code base + the test](https://github.com/search?q=repo%3AQiskit-Extensions%2Fcircuit-knitting-toolbox%20do_while&type=code)?

There are a bunch of instances of this migration at https://github.com/Qiskit/qiskit/pull/11448/files (scroll down to the very last change in that diff, for instance).

circuit_knitting/utils/simulation.py
```diff
@@ -170,5 +170,6 @@ def _run(
         **run_options,
     ):
         job = PrimitiveJob(self._call, circuits, parameter_values, **run_options)
-        job.submit()
+        # The public submit method was removed in Qiskit 1.0
+        (job.submit if hasattr(job, "submit") else job._submit)()
```
mrossinek, line 174: Seems a bit odd using a private method here. Then again, you do not really have an alternative with the `PrimitiveJob` interface.

Why exactly is this `ExactSampler` needed? I would assume that this works the same as `Sampler(..., options={"shots": None})`, does it not?

mrossinek (approved):
Just a minor thing but other than that these changes look good to me :+1:

circuit_knitting/utils/simulation.py
caleb-johnson: We needed to make `ExactSampler` because the reference implementation doesn't handle mid-circuit measurements, which is crucial for us.

Ya, this `(job.submit if ...)()` seemed to be the go-to solution for a couple of other repos as well. The public method didn't carry over from 0.46 to 1.0.

caleb-johnson: This is a good catch though, because they updated the [parent classes in the class signature](https://github.com/Qiskit/qiskit/blob/63c922e617709762837e2b1411d447f90de966b6/qiskit/primitives/sampler.py#L38) to include `PrimitiveJob` when `Sampler` started calling the `PrimitiveJob._submit` private method. We should do the same.

PR #478, opened 2024-02-07
Update ruff config

caleb-johnson:
`ruff` wants us to use the `lint` package in these two spots. This PR just removes a couple of warnings from our `lint` CI job

```
warning: The top-level linter settings are deprecated in favour of their counterparts in the `lint` section. Please update the following options in `.ruff.toml`:
  - 'ignore' -> 'lint.ignore'
  - 'per-file-ignores' -> 'lint.per-file-ignores'
```

### DIFF
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -1,5 +1,7 @@
 target-version = "py38"
-lint.ignore = [
+
+[lint]
+ignore = [
     "E501",  # line too long
 ]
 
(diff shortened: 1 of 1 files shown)


.ruff.toml
```diff
@@ -1,9 +1,9 @@
 target-version = "py38"
-ignore = [
+lint.ignore = [
```
garrison, line 2:
```suggestion

[lint]
ignore = [
```

garrison (dismissed):
Thanks, I have noticed these warnings too.  This LGTM.  I added one suggestion, which is optional.  (Hopefully it doesn't break anything if you apply it.)

PR #479, opened 2024-02-09
Remove entanglement forging

caleb-johnson:
Nothing in `Qiskit-Extensions` should rely on the old Qiskit application modules, so we must remove Entanglement Forging from this repository. We can add it to `qiskit-community` in a standalone repository if we want.


caleb-johnson:
Trying to track down what is causing this pandoc warning, I assume it's one of our notebook deps but not sure.

Other than that, this is ready for review. If we can get this and #480 removed, we can merge #475 and release. We could also clean up some deprecated packages as well.

PR #482, opened 2024-02-10
Drop dependency on qiskit-algorithms

garrison:
qiskit-algorithms is no longer needed as a dependency now that entanglement forging is gone.

This is a follow up to #479.


caleb-johnson (approved):
Awesome, thanks 🙌

PR #484, opened 2024-02-12
Remove qiskit.org links

Eric-Arellano:
These links were all working okay thanks to redirects. But it's better to use the correct link directly.


caleb-johnson (approved):
Great, thank you!

PR #498, opened 2024-03-06
Find cuts updates

caleb-johnson:
- Move `find_cuts` into `cutting_decomposition` module
- `find_cuts` returns metadata including cut types and locations and sampling overhead
- Adds a single test to get coverage of the function. Will add more tests
- Only accepts dicts to specify optimization and device settings

TODO:
- [Future PR] Write more tests
- [Future PR] Catch errors for 3+-qubit gates and tell users to decompose them
- [X]  Specify what settings are available in the docstring, since we won't have hyperlinks to the settings classes in the docs


caleb-johnson:
@Ibrahim-Shehzad , I'm going to add more tests and continue cleaning up a few things, but I wanted to merge this PR with updates to `find_cuts`

circuit_knitting/cutting/cutting_decomposition.py
```diff
@@ -260,3 +267,90 @@ def decompose_observables(
...
+
+    # Return metadata describing the cut scheme
+    sampling_overhead = 1
+    metadata: dict[str, Any] = {"cuts": []}
+    for i, inst in enumerate(circ_out.data):
+        if inst.operation.name == "qpd_2q":
+            sampling_overhead *= inst.operation.basis.overhead
+            metadata["cuts"].append(("Gate Cut", i))
+        if inst.operation.name == "cut_wire":
+            sampling_overhead *= 16
```
Ibrahim-Shehzad, line 352: So it looks like you're recomputing the sampling overhead here. I thought the cut finder already did that and that you could just access that directly. Or am I missing something here?

circuit_knitting/cutting/cutting_decomposition.py
```diff
@@ -260,3 +267,90 @@ def decompose_observables(
...
+    wire_cut_actions = []
+    gate_ids = []
+
+    opt_out = cast(DisjointSubcircuitsState, opt_out)
+    opt_out.actions = cast(list, opt_out.actions)
+    for action in opt_out.actions:
+        if action[0].get_name() == "CutTwoQubitGate":
+            gate_ids.append(action[1][0])
+        else:
+            # We only support three types of optimizer cuts
```
Ibrahim-Shehzad, line 324: In addition to `CutTwoQubitGate`, `CutRightWire`, and `CutLeftWire`, we actually also support `CutBothWires`.

circuit_knitting/cutting/cutting_decomposition.py
```diff
@@ -260,3 +267,90 @@ def decompose_observables(
...
+            assert action[0].get_name() in ("CutLeftWire", "CutRightWire")
```
Ibrahim-Shehzad, line 325: In addition to `CutTwoQubitGate`, `CutRightWire`, and `CutLeftWire`, we actually also support `CutBothWires`.

circuit_knitting/cutting/cutting_decomposition.py
caleb-johnson: Short answer: The QPD information isn't available in these CutWire objects

More in-depth answer:
[CutWire](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/ef510cae58f38d4c6090c33649796cbefc50c3e8/circuit_knitting/cutting/instructions/cut_wire.py#L19) is really just a dummy instruction indicating we need to allocate an ancilla qubit and do a [Move](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/ef510cae58f38d4c6090c33649796cbefc50c3e8/circuit_knitting/cutting/instructions/move.py#L19) operation at that location. It doesn't contain QPD information. After it is transformed to a ``Move`` operation, we can "cut" the Move operation, which creates two single qubit ``QPDGates`` with labels ["cut_move"](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/ef510cae58f38d4c6090c33649796cbefc50c3e8/circuit_knitting/cutting/qpd/instructions/qpd_gate.py#L144), one on either side of the wire cuts. These contain the QPD information.

I don't want to have to decompose these gates and allocate ancillas and all that, especially since we can hard-code the overhead for wire cuts trivially.

caleb-johnson: I thought you were asking specifically about the hard-coded value here. I see what you mean now. Let me see if we have the sampling overhead available in one of these variables somewhere 🧐

Ibrahim-Shehzad: I guess what I was getting at was that `opt_out.upper_bound_gamma()` would return to you the overall `gamma` for the cut scheme that is already computed internally. Isn't that what we are looking for in this context?

caleb-johnson: Yes, I realized what you were getting at after I typed all that out lol

caleb-johnson: OK I implemented this as you suggested

circuit_knitting/cutting/cutting_decomposition.py
caleb-johnson: I made a comment to follow up on this

test/cutting/test_cutting_decomposition.py
```diff
@@ -257,6 +259,19 @@ def test_partition_problem(self):
...
+            circuit = random_circuit(7, 6, max_operands=2, seed=1242)
+
+            cut_circ, metadata = find_cuts(
+                circuit, {"rand_seed": 111}, {"qubits_per_QPU": 4, "num_QPUs": 2}
+            )
+            cut_types = {cut[0] for cut in metadata["cuts"]}
+
+            assert len(metadata["cuts"]) == 2
+            assert {"Wire Cut", "Gate Cut"} == cut_types
+            assert np.isclose(metadata["sampling_overhead"], 127.06026169, atol=1e-8)
```
caleb-johnson, line 273: reverse order here
```suggestion
            assert np.isclose(127.06026169, metadata["sampling_overhead"], atol=1e-8)
```

circuit_knitting/cutting/cutting_decomposition.py
```diff
@@ -260,3 +267,105 @@ def decompose_observables(
...
+    circuit: QuantumCircuit,
+    optimization: dict[str, str | int],
+    constraints: dict[str, int],
+) -> tuple[QuantumCircuit, dict[str, Any]]:
+    """
+    Find cut locations in a circuit, given optimization settings and QPU constraints.
+
+    Args:
+        circuit: The circuit to cut
+        optimization: Settings for controlling optimizer behavior. Currently,
```
caleb-johnson, line 282: I'd like to explicitly state the optimization and constraint options the user has here before merging

circuit_knitting/cutting/cutting_decomposition.py
```diff
@@ -322,7 +322,11 @@ def find_cuts(
             gate_ids.append(action[1][0])
         else:
             # We only support three types of optimizer cuts
```
Ibrahim-Shehzad: Did you mean three types of _wire_ cuts? I would personally be inclined to either remove this comment or change it to something like ''We only support three actions in the context of wire cutting.''

circuit_knitting/cutting/cutting_decomposition.py
```diff
@@ -260,3 +267,105 @@ def decompose_observables(
...
+    wire_cut_actions = []
+    gate_ids = []
+
+    opt_out = cast(DisjointSubcircuitsState, opt_out)
+    opt_out.actions = cast(list, opt_out.actions)
+    for action in opt_out.actions:
+        if action[0].get_name() == "CutTwoQubitGate":
+            gate_ids.append(action[1][0])
+        else:
+            # We only support four types of optimizer cuts
```
caleb-johnson, line 324: @Ibrahim-Shehzad , I updated this commend when I added CutBothWires to the assert statement.

I normally just drop a comment above an assert statement explaining why I am making an assumption about a data structure. The code after this assert makes assumptions about how the Action is formatted, as we are assuming it's one of these 3 classes.

I like asserts for things like this because if we add a new class, instead of a strange index error, this assert would immediately flag our newly broken assumption.

Ibrahim-Shehzad: Right, that makes sense. I think I missed the update. I guess part of my concern was also whether ''types'' of optimizer cuts could be confused with cuts being "LO" vs "LOCC" but I think what you have is probably just fine. Thanks.

caleb-johnson: I will rephrase/improve this comment

Ibrahim-Shehzad: The updated comment looks good, thanks!

circuit_knitting/cutting/cutting_decomposition.py
```diff
@@ -260,3 +267,105 @@ def decompose_observables(
...
+        qubit_id = action[2][0][0] - 1
+        circ_out.data.insert(
+            inst_id + counter,
+            CircuitInstruction(
+                CutWire(), [circuit.data[inst_id + counter].qubits[qubit_id]], []
+            ),
+        )
+        counter += 1
+        if action[0].get_name() == "CutBothWires":
+            # qubit_id is 0 or 1. Flip it.
```
Ibrahim-Shehzad, line 364: This may be a CKT/Qiskit thing I am unfamiliar with but why do we do this flipping? Does it have to do with qiskit using little endian notation?

caleb-johnson: No, I'll clarify this in the comment. Basically I'm just saying that if this is a "CutBothWires", we have a second wire in our "Action" to cut and the gate-relative qubit for the second wire will be the opposite of whatever the first qubit was.

We know qubit_id is 0 or 1 here because we only support 2-qubit gates, so we just flip it and cut that one too. We don't actually need to read it from the structure, althought I suppose we could for clarity

caleb-johnson: OK we now read qubit_id2 from the data structure as well instead of this funky logic. It's more clear what is going on.

circuit_knitting/cutting/cutting_decomposition.py
```diff
@@ -260,3 +267,118 @@ def decompose_observables(
...
+        qubit_id = action[2][0][0] - 1
+        circ_out.data.insert(
+            inst_id + counter,
+            CircuitInstruction(
+                CutWire(), [circuit.data[inst_id + counter].qubits[qubit_id]], []
+            ),
+        )
+        counter += 1
+        if action[0].get_name() == "CutBothWires":
+            # There should be two wires specified in the action in this case
```
caleb-johnson, line 364: @Ibrahim-Shehzad , I think this is more clear.

Ibrahim-Shehzad: Yes, thanks!

circuit_knitting/cutting/cutting_decomposition.py
Ibrahim-Shehzad: Yes, thanks.

Ibrahim-Shehzad:
Everything looks good, thanks for putting this together.

PR #504, opened 2024-03-15
Add transpilation to cutqc evaluation

### LINKED ISSUE #501: evaluate_subcircuits issue after the 1st march 2024 update
#### Issue/Error:-
Sample code snippet:-
```python
#qc is any circuit
cuts = cut_circuit_wires(
        circuit=qc, method="manual", subcircuit_vertices=list(...)
    )
```
   --------------error is raised in below line------------------------
```
_subcircuit_instance_probabilities = evaluate_subcircuits(cuts, **kwargs)_
reconstructed_probabilities = reconstruct_full_distribution( qc, subcircuit_instance_probabilities, cuts, num_threads=1 )
```

(backend used is any real ibm hardware , and rest parameters are passed from the code)

After Cutting the circuit  when we pass the subcircuits and its complete path map generated i.e(cuts) by cut_circuit_wires function , 
evaluate_subcircuits is throwing an error shown below 

image

#### Reason for the Error:- 
[Update to Qiskit Runtime Primitives | IBM Quantum Platform](https://docs.quantum.ibm.com/announcements/product-updates/2024-02-14-qiskit-runtime-primitives-update)
As of 1 March 2024, circuits and observables need to be transformed to only use instructions supported by the system (referred to as instruction set architecture (ISA) circuits and observables) before being submitted to the Qiskit Runtime primitives.

#### Update Requested:-
Hence in evaluate_subcircuts function provided by cutting knitting toolbox , updates need to be done so that while running the circuits inside that function  it won't throw the above error.

garrison:
Fixes #501.

I've also replaced `service_args`, which appears nowhere else in the repository, with `service`, in the notebooks.

Unfortunately, even with this change, I see the following error when I try to run the first tutorial notebook on a hardware backend:

```
File ~/circuit-knitting-toolbox/circuit_knitting/cutting/cutqc/wire_cutting_evaluation.py:236, in run_subcircuits_using_sampler(subcircuits, sampler)
    234 for i, qd in enumerate(quasi_dists):
    235     probabilities = qd.nearest_probability_distribution()
--> 236     probabilities_out = np.zeros(2 ** subcircuits[i].num_qubits, dtype=float)
    238     for state in probabilities:
    239         probabilities_out[state] = probabilities[state]

ValueError: Maximum allowed dimension exceeded
```

I think this may be unrelated to the ISA change, but it would be nice to understand.

Also, right now the options to the pass-manager (in particular, optimization level = 1) are hardcoded.  It would be possible to make this configurable, but it would involve expanding a bunch of function signatures.

### DIFF
--- a/circuit_knitting/cutting/cutqc/wire_cutting_evaluation.py
+++ b/circuit_knitting/cutting/cutqc/wire_cutting_evaluation.py
@@ -24,6 +24,7 @@ from qiskit import QuantumCircuit
 from qiskit.converters import circuit_to_dag, dag_to_circuit
 from qiskit.circuit.library.standard_gates import HGate, SGate, SdgGate, XGate
 from qiskit.primitives import BaseSampler, Sampler as TestSampler
+from qiskit.transpiler.preset_passmanagers import generate_preset_pass_manager
 from qiskit_ibm_runtime import QiskitRuntimeService, Sampler, Session, Options
 
 
@@ -262,6 +263,12 @@ def run_subcircuits(
     if service is not None:
         session = Session(service=service, backend=backend_name)
         sampler = Sampler(session=session, options=options)
+        # Transpile circuits
+        backend = service.backend(session.backend())
+        pass_manager = generate_preset_pass_manager(
+            optimization_level=1, backend=backend
+        )
+        subcircuits = pass_manager.run(subcircuits)
     else:
         sampler = TestSampler(options=options)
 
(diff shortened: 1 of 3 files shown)


garrison:
Even if we can't figure out the `ValueError`, we _could_ just decide to merge this anyway.  All automated tests pass, and this is at least a step in the right direction now that Runtime only supports ISA circuits.

caleb-johnson:
Thank you for patching this! 🙌

caleb-johnson (approved):
Thanks for taking care of this! 🚀

PR #510, opened 2024-03-25
Split `qpd.py` code into three files

garrison:
The code currently in `qpd.py` implements three separate concepts, each of which has no dependency on the others.  For this reason, I believe it should be split into three modules.  The proposal here results in
- `weights.py` - includes `WeightType` and code for generating weights
- `decompose.py` - contains `decompose_qpd_instructions`
- `deompositions.py ` - implements `qpdbasis_from_instruction` and contains the `QPDBasis` decompositions for all two-qubit instructions we support

`test_qpd.py` _could_ be refactored along similar lines as well, but I wanted to start here to get some feedback on this.

The names decompose.py and decompositions.py and unfortunately similar.  I couldn't think of better names; in particularly, I really like calling the latter file decompositions, because that's what it entails.


caleb-johnson:
> The names decompose.py and decompositions.py and unfortunately similar. I couldn't think of better names; in particularly, I really like calling the latter file decompositions, because that's what it entails.

I think it's fair for a `decompose` module to pull its decompositions from a different module :).

caleb-johnson:
This is a breaking change, so I think we need a release note at least?

caleb-johnson:
> This is a breaking change, so I think we need a release note at least?

Maybe I'm stretching the definition of "breaking change". A user can still import from `qpd` in exactly the same way, but thhey could have been importing directly from the module as well, and it would have been valid, so I think we'd still consider it breaking

garrison:
> > This is a breaking change, so I think we need a release note at least?
> 
> Maybe I'm stretching the definition of "breaking change". A user can still import from `qpd` in exactly the same way, but thhey could have been importing directly from the module as well, and it would have been valid, so I think we'd still consider it breaking

Yeah, it really depends on how we want to approach this.  At the moment, I have followed qiskit's lead, where they only consider the top-level import (the one in `__init__.py`) to be stable.  See e.g. discussion at https://github.com/Qiskit/qiskit/pull/11635#issuecomment-1915535406.  We've never imported from `qpd.qpd` in our documentation, so I think we'd be safe just moving the code in that module.

But if we want to be more kind to anyone who is potentially using it anyway, we could do either or both of the following
- release note
- leave a `qpd.py` file that then imports from the new locations and shows a DeprecationWarning

caleb-johnson:
> > > This is a breaking change, so I think we need a release note at least?
> > 
> > 
> > Maybe I'm stretching the definition of "breaking change". A user can still import from `qpd` in exactly the same way, but thhey could have been importing directly from the module as well, and it would have been valid, so I think we'd still consider it breaking
> 
> Yeah, it really depends on how we want to approach this. At the moment, I have followed qiskit's lead, where they only consider the top-level import (the one in `__init__.py`) to be stable. See e.g. discussion at [Qiskit/qiskit#11635 (comment)](https://github.com/Qiskit/qiskit/pull/11635#issuecomment-1915535406). We've never imported from `qpd.qpd` in our documentation, so I think we'd be safe just moving the code in that module.
> 
> But if we want to be more kind to anyone who is potentially using it anyway, we could do either or both of the following
> 
> * release note
> * leave a `qpd.py` file that then imports from the new locations and shows a DeprecationWarning

I would personally forego the deprecation, but give the release note informing users they should be importing from the package level

caleb-johnson (dismissed):
Looks good! I'll let you decide how to handle deprecation/release note

PR #520, opened 2024-03-29
Add automated cut-finding module

Ibrahim-Shehzad:
Continuation of #471 .

PR #521, opened 2024-03-29
Construct test circuits in functions, not module

garrison:
Follow up to https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/520#discussion_r1544626447

These are still not _fixtures_, but at least they will only be run if the test is executed.


Ibrahim-Shehzad:
Yes, thanks, this looks good to me.

PR #527, opened 2024-04-02
Deprecate CutQC

Ibrahim-Shehzad:
Now that automated cut finding has been introduced (#520), we can go ahead and deprecate CutQC (#494).

PR #533, opened 2024-04-05
Update cutqc tutorials index with deprecated status

garrison:
This is a follow-up to #527.  Marking as a draft because it might be worth putting other updates of places we missed as we discover them.

#### To do

- [x] make sure the deprecation warning appears somehow in each cutqc tutorial (heh, even with #534 this is still necessary because the `%%capture` in the first `cutqc` import in each tutorial suppresses the `DeprecationWarning`).

### DIFF
--- a/docs/circuit_cutting/cutqc/index.rst
+++ b/docs/circuit_cutting/cutqc/index.rst
@@ -6,7 +6,7 @@ The ``cutqc`` module implements the wire cutting technique and
 automatic cut finding method described in the research paper
 `arXiv:2012.02333 <https://arxiv.org/abs/2012.02333>`_.  Historically,
 this was the original circuit cutting implementation in the Circuit
-Knitting Toolbox.  Going forward, this module is deprecated.  New users
+Knitting Toolbox.  Going forward, this module is deprecated.  Users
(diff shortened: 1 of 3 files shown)


docs/circuit_cutting/cutqc/index.rst
```diff
@@ -6,14 +6,8 @@ The ``cutqc`` module implements the wire cutting technique and
...
 this was the original circuit cutting implementation in the Circuit
-Knitting Toolbox.  Going forward, this module is considered legacy
-code that is supported, but new features are not expected.  New users
-of the toolbox should use the new circuit cutting interface, instead,
-unless they specifically need a feature that is not (yet) supported by
-the new code.  These features currently specific to ``cutqc`` include:
-
-- Reconstruction of probability distributions (rather than expectation values)
-- Automatic cut finding
+Knitting Toolbox.  Going forward, this module is deprecated.  New users
```
ibrahim-shehzad, line 9: Splitting hairs but since this applies to _all_ users, we may want to say: 
```suggestion
Knitting Toolbox.  Going forward, this module is deprecated.  Users
```

PR #537, opened 2024-04-08
Remove greedy recursion

### LINKED ISSUE #536: Max recursion limit exceeded when running cut-finder on deep circuits
For circuits with ~5000 non-local gates, such as those that appear in use cases of the QAOA, the default max recursion limit in python (3000) is exceeded. This is understandable because 5000 non-local gates involve around ~5000 recursive calls to the greedy BFS algorithm. 

Replace recursive calls to `greedy_best_first_search()` with a  different data structure, possibly a `heapq`.

ibrahim-shehzad:
This PR aims to resolve #536, by getting rid of recursion inside the `greedy_best_first_search` algorithm that is used to warm start the search performed by the cut-finder. I have replaced recursion with a simple `while` loop and have verified that the issues reported to us for large QAOA circuits (which led to #536) are now avoided. Since these circuits have large widths, the algorithm takes a while to finish running and any direct tests on these circuits would significantly slow down the overall testing process. This is why I have not added any new tests at this stage.


ibrahim-shehzad:
Looks like the `deprecate-cutqc` branch was renamed `remove-greedy-recursion` when I created a new branch for this locally. I think I may have rushed through something. Let me know if I should delete this and start afresh.

caleb-johnson:
@ibrahim-shehzad the diff for this PR includes a lot of stuff from the deprecation PR. Could we clean this up? This can happen if you have commits from a feature branch and then those commits are squashed when that branch is merged. Git doesnt know what to do so it keeps things in the diff that are not actual changes. I think you can just checkout the main versions of these files and then commit and push those files and it'll update the diff

caleb-johnson:
> With pieces of older git history showing up above, it looks like I may have (accidentally) created this branch locally from the deprecate-cutqc branch, instead of from main. Let me know if this is a problem and I should undo this and start over.

@ibrahim-shehzad This can happen if you have commits from a feature branch and then those commits are squashed when that branch is merged. Git doesnt know what to do so it keeps things in the diff that are not actual changes. I think you can just checkout the main versions of these files and then commit and push those files and it'll update the diff

circuit_knitting/cutting/cut_finding/cco_utils.py
```diff
@@ -138,23 +138,25 @@ def greedy_best_first_search(
...
-    )
-
-    if best[-1] is None:  # pragma: no cover
-        # This covers a rare edge case.
-        # We have so far found no circuit that triggers it.
-        # Excluding from test coverage for now.
-        return None
-
-    return greedy_best_first_search(best[-1], search_space_funcs, *args)
+    while True:
```
caleb-johnson, line 141: Do you have any thoughts about cases where this could run infinitely? What are the cases where `142` wouldn't evaluate to `True`?

Just a suggestion, but I like to bring return statements out of the loop if I can for clarity. I also like to include the condition in the while statement if I can to make it a little more clear when this thing should stop

```
>>> while not search_space_funcs.goal_state_func(...):
>>>     best = min(
>>>     ...
>>>     state = best[-1]
>>>     continue
>>> return state
```

Just a suggestion, it's up to you

caleb-johnson:
This looks nice, and it appears to implement the same logic. If you could just say something about what the conditions are when this loop should end, I'd like to consider if there are cases which could cause this to run forever.

circuit_knitting/cutting/cut_finding/cco_utils.py
ibrahim-shehzad: >Do you have any thoughts about cases where this could run infinitely? What are the cases where #142 wouldn't evaluate to True ever or for a long time?

So the way it works is that a circuit is represented as a list of gates and each gate in this list is ''visited,'' one after the other, whereby a decision on whether or not to cut and how to cut (wire vs gate) is made. Once all the gates have been visited (in the sense described above), the state obtained is a goal state. That is basically when the loop terminates. Therefore, the loop will run for longer when there is a large number of gates in the circuit.

caleb-johnson: OK, so this is really an O(N) loop under the hood, where N is the number of gates in the circuit. There isn't a condition where it would run infinitely or for an unknown amount of time that we can think of?

circuit_knitting/cutting/cut_finding/cco_utils.py
```diff
@@ -138,23 +138,25 @@ def greedy_best_first_search(
...
+        )
+        if best[-1] is None:  # pragma: no cover
+            # This covers a rare edge case.
+            # We have so far found no circuit that triggers it.
+            # Excluding from test coverage for now.
+            return None
+        else:
+            # avoid using recursion to stay clear of any risk of stack overflow.
+            state = best[-1]
+            continue
```
caleb-johnson, line 162: If you don't take my comment above, I think we can remove this continue. I don't think it does anything

circuit_knitting/cutting/cut_finding/cco_utils.py
```diff
@@ -138,23 +138,25 @@ def greedy_best_first_search(
...
-
-    if best[-1] is None:  # pragma: no cover
-        # This covers a rare edge case.
-        # We have so far found no circuit that triggers it.
-        # Excluding from test coverage for now.
-        return None
-
-    return greedy_best_first_search(best[-1], search_space_funcs, *args)
+    while True:
+        if search_space_funcs.goal_state_func(state, *args):
```
caleb-johnson, line 142: It seems we don't expect this to be `None`. That is the rare case below. So in the case where we aren't ready to return yet, we expect this to be `[]` or what here? Maybe we can explicitly state it in the `if` statement.

caleb-johnson: OK sorry I got a little mixed up. Disregard above.

circuit_knitting/cutting/cut_finding/cco_utils.py
ibrahim-shehzad: I think that's right, yes.

ibrahim-shehzad: >Just a suggestion, but I like to bring return statements out of the loop if I can for clarity. I also like to include the condition in the while statement if I can to make it a little more clear when this thing should stop.

 I think this sounds good. Will work on this.

ibrahim-shehzad: Updated: [dc43c78](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/537/commits/dc43c789526fa975fb75819f06e9a1ce70dce6aa)

circuit_knitting/cutting/cut_finding/cco_utils.py
```diff
@@ -138,23 +138,23 @@ def greedy_best_first_search(
...
+        )
+        if best[-1] is None:  # pragma: no cover
+            # This covers a rare edge case.
+            # We have so far found no circuit that triggers it.
+            # Excluding from test coverage for now.
+            return None
+        else:
+            # avoid using recursion to stay clear of any risk of stack overflow.
+            state = best[-1]
+            continue
```
garrison, line 159: Following [Caleb's comment](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/537/files#r1557810322), I suggest

```suggestion
        state = best[-1]
```

ibrahim-shehzad: Thanks, I have made this change here: [183c13a](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/537/commits/183c13a2d91fc9b40b45f43db61b1106d6eb9d8e)

circuit_knitting/cutting/cut_finding/cco_utils.py
```diff
@@ -138,23 +138,20 @@ def greedy_best_first_search(
...
+                )
+            ],
+            default=(None, None, None),
+        )
+        if best[-1] is None:  # pragma: no cover
+            # This covers a rare edge case.
+            # We have so far found no circuit that triggers it.
+            # Excluding from test coverage for now.
+            return None
+        state = best[-1]
```
caleb-johnson, line 156: I prefer using new varnames (e.g. `state_out`) to modifying input varnames, unless I intend on modifying the input variable in-place in memory, in which case I would document it in the docstring. You simply reassign `state` to a new data structure in memory, which is fine, but I do try to avoid messing with the input variables altogether as a safety measure against accidentally modifying an input data structure.

What you have here is fine and safe, so it's up to you what you prefer.

circuit_knitting/cutting/cut_finding/cco_utils.py
```diff
@@ -138,23 +138,20 @@ def greedy_best_first_search(
...
+                    search_space_funcs.next_state_func(state, *args)
+                )
+            ],
+            default=(None, None, None),
+        )
+        if best[-1] is None:  # pragma: no cover
+            # This covers a rare edge case.
+            # We have so far found no circuit that triggers it.
+            # Excluding from test coverage for now.
+            return None
```
caleb-johnson, line 155: If you take my below comment you can just say:

```
state_out = None
break
```
and you won't have any returns in the loop anymore. This is another style choice, but as loops grow bigger and more return statements are added, it can become messy, so I've made a habit of pushing returns out of loops and setting appropriate variables within the loop.

Readability/understandability is the primary goal, so whatever you think is best here as well

caleb-johnson (approved):
This looks good to me. A couple of small style suggestions, but it's fine to merge as-is IMO. Thanks!

circuit_knitting/cutting/cut_finding/cco_utils.py
garrison: I actually tend to prefer a `return` in the loop over a `break` statement, so Ibrahim, it is truly up to you; as Caleb says, readability/understandability is the primary goal.

PR #540, opened 2024-04-09
Add github button to docs homepage

caleb-johnson:
If I end up at a software package's docs page before their Github repo, it is nice to be able to get directly to the Github repo from there. I find github repos can often be more difficult to find on search engines than docs pages

We have a small button which shows our stars and links users to Github. I think that is great, and I think we can also just add a button to our toolbar that takes people there.

This is personal preference, but this is something I personally think is worth having if your homescreen toolbar isn't already crowded

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -53,6 +53,7 @@ Contents
 
   API References <apidocs/index>
   Release Notes <release-notes>
+  Github <https://github.com/Qiskit-Extensions/circuit-knitting-toolbox>
 
 .. Hiding - Indices and tables
    :ref:`genindex`
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -53,7 +53,7 @@ Contents
 
   API References <apidocs/index>
   Release Notes <release-notes>
-  Github <https://github.com/Qiskit-Extensions/circuit-knitting-toolbox>
+  GitHub <https://github.com/Qiskit-Extensions/circuit-knitting-toolbox>
 
 .. Hiding - Indices and tables
    :ref:`genindex`


garrison:
> We have a small button which shows our stars and links users to Github

I literally just clicked on that two minutes ago :laughing:.

But yeah, it's one of those things where if you know, you know, and if you don't know, then ... you don't.

docs/index.rst
```diff
@@ -53,6 +53,7 @@ Contents
 
   API References <apidocs/index>
   Release Notes <release-notes>
+  Github <https://github.com/Qiskit-Extensions/circuit-knitting-toolbox>
```
garrison, line 56:
```suggestion
  GitHub <https://github.com/Qiskit-Extensions/circuit-knitting-toolbox>
```

garrison:
Not sure if you noticed it, but there is also a linked GitHub logo in the footer of each page of the Sphinx build.

I mention it because it is pretty subtle, but I think having this in the sidebar might be subtle too.  Still an improvement, though.

caleb-johnson:
> Not sure if you noticed it, but there is also a linked GitHub logo in the footer of each page of the Sphinx build.
> 
> I mention it because it is pretty subtle, but I think having this in the sidebar might be subtle too. Still an improvement, though.

Ahh I forgot about those. I think you may have put those in first, before the "Stars" shield.

caleb-johnson:
> ckt-github-link
> 
> I feel like it may not belong under "References". Would you have noticed it if it were instead right below Installation Instructions? Or, maybe we could get it at the bottom with some whitespace separating it from the References, but IIRC that would be tricky -- the margin mainly comes from the header.

I think it fits naturally into all 3 places you suggested TBH. You could be "referencing" the implementation, or it could be seen as a helpful introductory link, along with the other intro stuff. Putting it below with some padded whitespace is fine too, but as we add more tools to CKT, whitespace may get scarce

garrison (approved):
> You could be "referencing" the implementation

OK, you've convinced me.

caleb-johnson:
> OK, you've convinced me.

lol 🙌

PR #541, opened 2024-04-10
Adjust docs artifact filename

garrison:
I've been referencing the documentation artifact a lot lately.   It'd be nice to have better filenames to work with.

Formerly: `html_docs.zip` extracted directly to the current directory.

With this PR: `ckt_html_docs.zip` extracts everything to a single subdirectory called `ckt_html_docs`.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -37,12 +37,18 @@ jobs:
         shell: bash
         run: |
           tox -edocs
+      - name: Prepare docs artifact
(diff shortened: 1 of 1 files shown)


garrison:
> With this PR: `ckt_html_docs.zip` extracts everything to a single subdirectory called `ckt_html_docs`.

I checked the resulting artifact and verified that this is indeed the behavior under this PR.

Eric-Arellano (approved):

PR #544, opened 2024-04-12
Enable pylint's `dangerous-default-value` check

garrison:
When I ran pylint in its default configuration, I noticed a few `dangerous-default-value` warnings in the new cut finder.  I think it's worth adding this to our pylint configuration and fixing the existing issues in the code base.  (Even though users aren't supposed to directly call this code, adding this to our config allows us to prevent similar dangerous defaults in places we _do_ care about.)

### DIFF
--- a/test/cutting/cut_finding/test_circuit_interfaces.py
+++ b/test/cutting/cut_finding/test_circuit_interfaces.py
@@ -68,7 +68,7 @@ class TestCircuitInterface:
         assert max_wire_cuts_gamma(7) == 2
 
         # Assign by hand a different qubit mapping by specifiying init_qubit_names.
-        circuit_converted = SimpleGateList(trial_circuit, ("q0", "q1"))
+        circuit_converted = SimpleGateList(trial_circuit, ["q0", "q1"])
         assert circuit_converted.qubit_names.item_dict == {"q0": 0, "q1": 1}
         assert circuit_converted.circuit == [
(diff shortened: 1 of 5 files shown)


ibrahim-shehzad (approved):
Interesting! I had to look at the examples given [here](https://docs.python.org/3/tutorial/controlflow.html#default-argument-values) to understand what the potential danger was but this makes sense now. This is super important. I made a minor modification to one of the tests and a docstring, just to make sure everything was consistent. This looks good to me overall, thanks

test/cutting/cut_finding/test_circuit_interfaces.py
```diff
@@ -68,7 +68,7 @@ def test_circuit_conversion(self):
         assert max_wire_cuts_gamma(7) == 2
 
         # Assign by hand a different qubit mapping by specifiying init_qubit_names.
-        circuit_converted = SimpleGateList(trial_circuit, ["q0", "q1"])
+        circuit_converted = SimpleGateList(trial_circuit, ("q0", "q1"))
```
garrison, line 71: I think this test could stay as it was, actually.  It's okay to pass a `list`; it's just not okay when the default function argument is a `list`.

ibrahim-shehzad: Yes but I just figured this would be "more" consistent?

ibrahim-shehzad: Your call though.

PR #556, opened 2024-04-22
Remove transpilations inside subexperiment generation

### LINKED ISSUE #529: Profile subexperiment generation
With an increased focus on utility-scale experiments, we need to profile our subexperiment generation to ensure there are no easily-correctable inefficiencies. I have noticed that for cutting schemes involving lower-rotation gates, subexperiment generation can take a very long time when doing a full sampling, even for modestly-sized circuits.

### LINKED ISSUE #530: Time a transpilation routine for a utility-scale set of subexperiments
Record the time it takes to do a standard transpilation routine on a utility-scale batch of subexperiments. We should allow for transpilation of the QPD circuit and try to avoid thousands of copies and transpilations at the subexperiment level, but it isn't clear how much of an improvement this will make on the runtime of a utility workflow. Let's benchmark this process to establish a baseline.

caleb-johnson:
Fixes #529 
Fixes #530 

The copies used in the transpilation passes in subexperiment generation start becoming a bottleneck for larger experiments. This PR removes the transpilation in favor of modifying the QuantumCircuit objects directly and in-place. This is more in-line with the rest of our workflow and scales much better in that context.
subexp_gen

- [x] Release note


caleb-johnson:
subexp_gen


@garrison @ibrahim-shehzad I do think we should move away from DAGCircuit in our workflow until we decide to move entirely to it. This chart shows a pretty dramatic time cost as experiments scale up if we continue using transpiler passes to clean up resets.

caleb-johnson:
Something funny going on with the reno linter in CI for this PR hmmm

caleb-johnson:
> Something funny going on with the reno linter in CI for this PR hmmm

Docs build is broken on main branch now as well. Opened #557 to address.

caleb-johnson:
The chart was made using Tutorial 1 on the All-Z observable. So 2 partitions, 1 observable.

circuit_knitting/cutting/cutting_experiments.py
```diff
@@ -389,3 +371,80 @@ def _get_pauli_indices(cog: CommutingObservableGroup) -> list[int]:
...
+    for i, inst in enumerate(circuit.data):
+        qargs = [circuit.find_bit(q).index for q in inst.qubits]
+        if inst.operation.name == "reset":
+            if resets[qargs[0]]:
+                remove_ids.append(i)
+            else:
+                resets[qargs[0]] = True
+        else:
+            for q in qargs:
+                resets[q] = False
```
ibrahim-shehzad, line 395: I may be missing something here but does the code block in lines 394-395 need to be there given that in line 382, `resets` is initialized to all its elements being `False`?

caleb-johnson: Yes, because a qubit could have been flipped to `True` at some point. This line is saying that we have a non-reset instruction and to flip all of the associated qargs back to False (indicating that the most recent instruction on that qubit was not a reset)

ibrahim-shehzad: I see, that makes sense, thanks.

ibrahim-shehzad (dismissed):
The speedup is really impressive! Thanks for putting this all together, it looks great to me.

releasenotes/notes/subexperiment-gen-speedup-41a4e8679353d1d9.yaml
```diff
@@ -0,0 +1,4 @@
+---
+upgrade:
```
garrison, line 2:
```suggestion
fixes:
```
It's not really something you need to know when upgrading; it's just that performance is all around better now.  IMO this is the more appropriate section.

circuit_knitting/cutting/cutting_experiments.py
```diff
@@ -389,3 +371,80 @@ def _get_pauli_indices(cog: CommutingObservableGroup) -> list[int]:
...
+
+
+def _remove_final_resets(circuit: QuantumCircuit, inplace=True) -> QuantumCircuit:
+    """Remove resets if they are the final instruction on a qubit."""
+    if not inplace:  # pragma: no cover
+        circuit = deepcopy(circuit)
+
+    # Keep up with whether we are at the end of a qubit
+    # We iterate in reverse, so all qubits begin in the "end" state
+    qubit_ended = set([i for i in range(circuit.num_qubits)])
```
garrison, line 435:
```suggestion
    qubit_ended = set(range(circuit.num_qubits))
```

circuit_knitting/cutting/cutting_experiments.py
```diff
@@ -389,3 +371,80 @@ def _get_pauli_indices(cog: CommutingObservableGroup) -> list[int]:
...
     return pauli_indices
+
+
+def _consolidate_resets(circuit: QuantumCircuit, inplace=True) -> QuantumCircuit:
+    """Consolidate redundant resets into a single reset."""
+    if not inplace:  # pragma: no cover
+        circuit = deepcopy(circuit)
+
+    # Keep up with whether the previous instruction on a given qubit was a reset
+    resets = {i: False for i in range(circuit.num_qubits)}
```
garrison, line 382: should potentially be a bit faster
```suggestion
    resets = [False for i in range(circuit.num_qubits)]
```

circuit_knitting/cutting/cutting_experiments.py
```diff
@@ -389,3 +371,80 @@ def _get_pauli_indices(cog: CommutingObservableGroup) -> list[int]:
     if not pauli_indices:
         pauli_indices = [0]
     return pauli_indices
+
+
+def _consolidate_resets(circuit: QuantumCircuit, inplace=True) -> QuantumCircuit:
+    """Consolidate redundant resets into a single reset."""
+    if not inplace:  # pragma: no cover
+        circuit = deepcopy(circuit)
```
garrison, line 379: why not just `qc.copy()`? which is what we have done in other places in case of `if not inplace`

circuit_knitting/cutting/cutting_experiments.py
```diff
@@ -389,3 +371,80 @@ def _get_pauli_indices(cog: CommutingObservableGroup) -> list[int]:
...
+        qargs = [circuit.find_bit(q).index for q in inst.qubits]
+        if inst.operation.name == "reset":
+            if qargs[0] in qubit_ended:
+                remove_ids.append(num_inst - 1 - i)
+        else:
+            for q in qargs:
+                qubit_ended.discard(q)
+
+    for i in sorted(remove_ids, reverse=True):
+        del circuit.data[i]
```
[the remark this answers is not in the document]
caleb-johnson: Yes I had wondered if there was some technique for removing many items for a list at the same time, but I couldn't find anything that looked to be exactly what I imagined

circuit_knitting/cutting/cutting_experiments.py
```diff
@@ -389,3 +370,80 @@ def _get_pauli_indices(cog: CommutingObservableGroup) -> list[int]:
     if not pauli_indices:
         pauli_indices = [0]
     return pauli_indices
+
+
+def _consolidate_resets(circuit: QuantumCircuit, inplace=True) -> QuantumCircuit:
```
garrison, line 375: Slight improvement would be
```suggestion
def _consolidate_resets(circuit: QuantumCircuit, inplace: bool = True) -> QuantumCircuit:
```
and likewise for the other functions.  (maybe this will need black to run, too -- not sure)

garrison (dismissed):
Looks great, thanks!

Issue #557, opened 2024-04-22
Docs build failing on main branch


caleb-johnson:
tox -e docs
pkg: install_requires> python -I -m pip install hatchling
.pkg: _optional_hooks> python /opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/site-packages/pyproject_api/_backend.py True hatchling.build
.pkg: get_requires_for_build_sdist> python /opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/site-packages/pyproject_api/_backend.py True hatchling.build
.pkg: freeze> python -m pip freeze --all
.pkg: hatchling==1.24.2,packaging==24.0,pathspec==0.[12](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/actions/runs/8791924164/job/24127118171?pr=556#step:6:13).1,pip==24.0,pluggy==1.5.0,setuptools==69.5.1,tomli==2.0.1,trove-classifiers==2024.4.10,wheel==0.43.0
.pkg: build_sdist> python /opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/site-packages/pyproject_api/_backend.py True hatchling.build
docs: install_package_deps> python -I -m pip install 'Sphinx>=3.0.0' 'cplex>=22.1.0.0; python_version < "3.12" and platform_machine != "arm64"' 'docplex>=2.23.222; python_version < "3.12" and platform_machine != "arm64"' 'ipykernel>=4.5.1' ipywidgets 'jupyter-sphinx>=0.3.2' matplotlib 'nbsphinx>=0.8.8' 'numpy>=1.23.0' pylatexenc 'qiskit-aer>=0.14.0.1' 'qiskit-ibm-runtime>=0.23.0' 'qiskit-sphinx-theme<2,>=1.14.0' 'qiskit<2.0,>=1.0.0' 'reno>=3.4.0' 'rustworkx>=0.14.0' 'scipy>=1.5.2' 'sphinx-autodoc-typehints>=1.12.0' 'sphinx-copybutton>=0.5.0' sphinx-reredirects
docs: install_package> python -I -m pip install --force-reinstall --no-deps /home/runner/work/circuit-knitting-toolbox/circuit-knitting-toolbox/.tox/.tmp/package/1/circuit_knitting_toolbox-0.7.0.tar.gz
docs: freeze> python -m pip freeze --all
[...]

PR #559, opened 2024-04-23
No worktrees

caleb-johnson:
Fixes https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/issues/557

There is a [problem with dulwich](https://github.com/jelmer/dulwich/issues/1285) (used by reno) and git worktrees. This PR implements a temporary fix, spelled out [here](https://github.com/GitTools/actions/issues/1115#issuecomment-2070736857).

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -28,6 +28,11 @@ jobs:
           pip install tox
           sudo apt-get update
           sudo apt-get install -y pandoc
+      - name: Temporary workaround for GitVersion
+        shell: bash
+        run: |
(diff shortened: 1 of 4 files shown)


garrison:
Thank you for tracking this down.  If I understand correctly, we should be able to revert this once upstream is fixed; is that correct?

garrison (dismissed):

caleb-johnson:
> Thank you for tracking this down.  If I understand correctly, we should be able to revert this once upstream is fixed; is that correct?  

Yes it seems to me the fix was very simple and should be there next dulwich release

Issue #515, opened 2024-03-27
Timeout during `generate_cutting_experiments` in `how_to_generate_exact_sampling_coefficients.ipynb`


garrison:
Now that #514 is merged, we can actually see which notebook hangs intermittently on CI.

```
=================================== FAILURES ===================================
_ /home/runner/work/circuit-knitting-toolbox/circuit-knitting-toolbox/docs/circuit_cutting/how-tos/how_to_generate_exact_sampling_coefficients.ipynb _
A cell timed out while it was being executed, after 300 seconds.
The message was: Cell execution timed out.
Here is a preview of the cell contents:
-------------------
subexperiments, coefficients = generate_cutting_experiments(
    circuits=subcircuits,
    observables=subobservables,
    num_samples=np.inf,
)
coefficients
-------------------

------------------------------ Captured log call -------------------------------
ERROR    traitlets:client.py:845 Timeout waiting for execute reply (300s).
```

Full log at:
- https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/actions/runs/8456649627/job/23166889295?pr=471 (ubuntu-latest, 3.8)
- https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/actions/runs/8475888125/job/23224662100#step:6:153 (ubuntu-latest, 3.8)
- https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/actions/runs/8511573980/job/23311448303#step:6:162 (ubuntu-latest, 3.8)
- https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/actions/runs/8513967433 (ubuntu-latest, 3.8)
- https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/actions/runs/8542748098/job/23405055909#step:6:163 (ubuntu-latest, 3.8)
- https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/actions/runs/8585209021/job/23526356875 (ubuntu-latest, 3.8)
- https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/actions/runs/8604410132/job/23578451818?pr=538 (ubuntu-latest, 3.8)

garrison:
I keep on adding to the list above, but so far this only seems to happen on Python 3.8 on Ubuntu.

garrison:
I noticed a warning when running the test suite under Python 3.12 that may be a clue.

```
[...]

PR #560, opened 2024-04-24
Test on ARM macOS runners

garrison:
~~This is due to python 3.8 no longer being available on the macos-latest CI runner.~~

This ~also~ allows us to now perform CI on ARM chips on macOS.

This also restores the ubuntu minimum version to 3.8 so that we test on 3.8 somewhere. https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/issues/515#issuecomment-2073378582

~~We might see intermittent CI failures again until #556 is merged (expected to be very soon).~~

### DIFF
--- a/.github/workflows/test_latest_versions.yml
+++ b/.github/workflows/test_latest_versions.yml
@@ -23,7 +23,7 @@ jobs:
         python-version: ["3.8", "3.12"]
         include:
           - os: macos-latest
-            python-version: "3.11"
+            python-version: "3.12"
           - os: windows-latest
             python-version: "3.10"
     steps:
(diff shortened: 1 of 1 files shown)


garrison:
I merged $561, which is an alternative to this.

If we want to take this route, we will have to disable the cutqc tests when cplex is not available.  This should be fine, since cutqc is deprecated now anyway.

garrison:
This is related to https://github.com/Qiskit/qiskit/pull/12102.  We might want to wait on that before merging.  But otherwise this is all done.

garrison:
https://github.com/Qiskit/qiskit/pull/12102 is now merged.  If testing on ARM causes us any intermittent problems before Qiskit 1.1 is released (I don't expect it to), then I can make a follow-up PR so that ARM testing uses qiskit `main` or `stable/1.1` until qiskit 1.1 is released.

garrison:
Merging this since Qiskit 1.1 will be released tomorrow.

PR #565, opened 2024-04-25
Early terminate from loops when generating cutting experiments

garrison:
This is a follow-up to #556.

### DIFF
--- a/circuit_knitting/cutting/cutting_experiments.py
+++ b/circuit_knitting/cutting/cutting_experiments.py
@@ -409,7 +409,7 @@ def _remove_resets_in_zero_state(
         circuit = circuit.copy()
 
     # Keep up with which qubits have at least one non-reset instruction
-    active_qubits = set()
+    active_qubits: set[int] = set()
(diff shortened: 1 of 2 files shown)


caleb-johnson (approved):
Thank you! This is something I intended to do at some point, and it slipped my mind 🚀

circuit_knitting/cutting/cutting_experiments.py
```diff
@@ -409,7 +409,7 @@ def _remove_resets_in_zero_state(
         circuit = circuit.copy()
 
     # Keep up with which qubits have at least one non-reset instruction
-    active_qubits = set()
+    active_qubits: set[int] = set()
```
caleb-johnson, line 412: Is this for clarity or was there some other reason?

garrison: The linter complained once I made the early termination change until I added this annotation.  (I even reverted back to your original, just to double check the linter was fine with it on my machine, too.)

PR #567, opened 2024-04-26
Set package `__version__` attribute

garrison:
With this:

```
$ python -c "import circuit_knitting; print(circuit_knitting.__version__)"
0.7.1
```

This essentially follows what we used to do at https://github.com/qiskit-community/quantum-prototype-template/blob/main/prototype_template/__init__.py.

### DIFF
--- a/circuit_knitting/__init__.py
+++ b/circuit_knitting/__init__.py
@@ -10,3 +10,15 @@
 # that they have been altered from the originals.
 
 """Main Circuit Knitting Toolbox public functionality."""
+
+from importlib.metadata import version, PackageNotFoundError
+
+
+try:
+    __version__ = version("circuit-knitting-toolbox")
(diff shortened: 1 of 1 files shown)


circuit_knitting/__init__.py
```diff
@@ -20,4 +20,4 @@
     # Package is not installed
     pass
 
-__all__ = []
+__all__: list[str] = []
```
caleb-johnson, line 24: I was surprised the linter flagged this , do we do this other places?

caleb-johnson (approved):

circuit_knitting/__init__.py
garrison: It's only because it's an empty list that the annotation is needed. I wanted to prevent the two imports from being re-exported, which is why I included this line.

PR #568, opened 2024-04-26
Attempt at implementing optimal cutting of a single gate

garrison:
This is my attempt to implement the optimal cutting of a _single_ two-qubit gate following https://arxiv.org/abs/2312.11638.  I have written out the local operations in terms of the operations in https://arxiv.org/abs/2006.11174 and combined like terms, and everything works.  However, my new test that `1 <= gamma <= 7` fails, so I think this may be equivalent to the previous code in terms of overhead and therefore not actually optimal :disappointed:.  At the moment, I am not sure if it is going to be possible to implement the local channels in the new paper given only single-qubit unitaries and projective measurements (i.e., without projective measurements on ancilla qubits).  I will keep thinking about these channels and ask around.

### DIFF
--- a/circuit_knitting/cutting/qpd/decompositions.py
+++ b/circuit_knitting/cutting/qpd/decompositions.py
@@ -13,6 +13,7 @@
 
 from __future__ import annotations
 
+import math
 from collections.abc import Sequence, Callable
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


garrison:
There is a revised version of the paper out, at https://arxiv.org/abs/2312.11638v3, which describes the operations in a new appendix.  The general operation does indeed require an ancilla qubit on each side of the cut.

PR #578, opened 2024-05-03
RFC: Use hatch-vcs to determine project's version from git

garrison:
This uses [hatch-vcs](https://github.com/ofek/hatch-vcs) to set the version automatically based on the most recent git tag, instead of having it specified and bumped manually in `pyproject.toml`.  Two benefits I can think of

- One fewer step (line to update) when we do a release
- Each git commit will be given a unique version string.  For instance, with this PR, _only_ when the version is equivalent to the `0.7.1` tag will it be returned as `0.7.1` precisely.  Otherwise, it will be a dev version string including the hash of the commit in use (such as the following repl demonstration for the current commit in this PR).  This will give us much better version information if we want to know precisely what (possibly development) version of ckt a user has installed.
  ```
  >>> import circuit_knitting
  >>> circuit_knitting.__version__
  '0.7.2.dev2+g4e93e75'
  ```

Downside:
- adds a build dependency on hatch-vcs

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,10 +1,10 @@
 [build-system]
-requires = ["hatchling"]
+requires = ["hatchling", "hatch-vcs"]
 build-backend = "hatchling.build"
 
 [project]
 name = "circuit-knitting-toolbox"
-version = "0.7.1"
+dynamic = ["version"]
 description = "A software prototype for a circuit knitting toolbox which connects user applications with runtime primitives"
 readme = "README.md"
 license = {file = "LICENSE.txt"}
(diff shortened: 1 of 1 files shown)


garrison:
If we go this route, we'll have to think carefully about what version information should be shown by Sphinx on the stable doc build.  Currently, it's going to pick up some of this added stuff in the version specifier, but this can be fixed by changing the following few lines:

https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/fe2da9a11b123038227b720edfb59b7ef929f63d/docs/conf.py#L33-L36

My biggest problem with hatch-vcs is that it always seems to think the next release is a patchlevel one, and there seems to be no way to configure this (at least, no way that I've found so far).  For instance, it thinks the next release on `main` is going to be `0.7.2`, not `0.8.0`.

PR #579, opened 2024-05-08
Improve docs and tests of `observables_restricted_to_subsystem`

garrison:
- mention that phases are discarded
- add `list[Pauli]` test case


circuit_knitting/utils/observable_grouping.py
```diff
@@ -45,14 +45,16 @@ def observables_restricted_to_subsystem(
     A :class:`~qiskit.quantum_info.PauliList` will be returned if a :class:`~qiskit.quantum_info.PauliList` is provided; otherwise,
     a ``list[Pauli]`` will be returned.
 
+    Any phase information will be discarded, consistent with the standard behavior when slicing a Pauli.
```
caleb-johnson, line 48: You could mention the other function(s) that has this behavior, since a user may not immediately know what all this "slicing a Pauli" means exactly.

test/utils/test_observable_grouping.py
```diff
@@ -77,6 +77,11 @@ def test_most_general_observable(self):
             PauliList(["XX", "YY", "ZZ"]),
             PauliList.from_symplectic(np.zeros((3, 0)), np.zeros((3, 0))),
         ),
+        (
+            (2, 1),
```
caleb-johnson, line 81: So flipping the order of IDs, flips the order of the Paulis..I'm sure this is intended, but worth staring at for a second 👀

circuit_knitting/utils/observable_grouping.py
garrison: Here's an example of what I mean that the phase gets discarded:
```python
>>> from qiskit.quantum_info import Pauli
>>> Pauli("iXZX")[:2]
Pauli('ZX')
```
It's the act of slicing, not really a function, but I will try to think of a better explanation that doesn't require somebody to know Python or programming jargon.

caleb-johnson: Ohhh I see. Actual Python slicing. No, this is fine. Thanks

test/utils/test_observable_grouping.py
garrison: Yeah, I wanted to make sure there was an actual test where things are meant to get re-ordered.

PR #581, opened 2024-05-08
Improve validation in the cutting reconstruction function

garrison:
- I took the first few validations and moved them into the following `if` block, since the conditionals are a bit redundant.  This looked better to me once I wrote everything out.
- The `dict` can really be any `Mapping`, so I use that broader class when calling `isinstance`.  At the same time, I've left `dict` in the type annotations for the sake of users who are reading the docs but don't know what a "`Mapping`" is.
- The thing that prompted this was the last chunk, which validates the number of subexperiments.  The lack of this error led to an experience where @ibrahim-shehzad was getting "wrong" results.  (It turned out, the `observables` were different than the ones used when generating the subexperiments; having this check will at least make sure the _number_ of observables makes sense.)

#### Remaining action items

- [x] Add tests to bring coverage to 100%


circuit_knitting/cutting/cutting_reconstruction.py
```diff
@@ -75,40 +75,62 @@ def reconstruct_expectation_values(
...
+        }
         expvals = np.zeros(len(observables))
 
-    else:
+    elif isinstance(observables, Mapping):
+        if not isinstance(results, Mapping):
+            raise ValueError(
+                "If observables is a dictionary, results must also be a dictionary."
+            )
+        if observables.keys() != results.keys():
```
caleb-johnson, line 100: This came about because utility-scale workloads may be run in chunks, saved to disk, and recombined, which means there could very easily be errors during that recombination?

caleb-johnson (approved):
Looks good, thanks! 👍

circuit_knitting/cutting/cutting_reconstruction.py
garrison: That is indeed one motivation.  Mostly I just wanted to add this as an additional sanity check when passing over this code, since the below lines implicitly make this assumption.

PR #585, opened 2024-05-10
Avoid exploring extraneous minima in the cut-finder search space

ibrahim-shehzad:
A certain step in the  cut-finder explores multiple minima of the cost function. This step, however, is only relevant when we have LOCC or LO blackbox/parallel gate cut QPD assignments. As such, it is unnecessary for the LO circuit cutting that we currently support. Disabling the search for multiple minima also helps speed up the performance of the cut finder. This was especially relevant for certain QAOA circuits that were reported to us, which involved multiple Rzz gates with $\theta =0$ (in which case the cost of cutting each of these gates is 1). Telling the cut finder to stop at the first minimum stops it from exploring an evergrowing list of states (since no pruning of states can take place when the cost of each additional gate cut just gives us a multiplicative factor of 1 to the overall cost). This PR aims to fix all of this, simply by setting the default value of certain `stop_at_first_min` flags to `True`.


circuit_knitting/cutting/cut_finding/best_first_search.py
```diff
@@ -149,6 +149,8 @@ class BestFirstSearch:
 
     ``stop_at_first_min`` (Boolean) is a flag that indicates whether or not to
     stop the search after the first minimum-cost goal state has been reached.
+    In the absence of any QPD assignments, it always makes sense to stop once
```
garrison, line 152: What does "qpd assignments"mean?

ibrahim-shehzad: Thanks for raising this. So the way Ed's overall code was designed was that in the first stage, it tries to find locations for all the cuts (assuming everything is just LO, since that provides a cost (overhead) upperbound) and then, in the second stage, assigns the actual QPD's (where the user can chose if it's instead LOCC, blackbox LO, etc.). The idea behind giving the user this choice was that being able to pick between LO, LOCC, and blackbox LO etc. can depend on the resources that are available (e.g whether or not CC is possible and whether or not you have access to ancillas etc). One should also in principle be able to make a different choice for each cut. In this situation, not all sets of cuts found in the first step are created equal; some may be more suitable than others.

Now when you're only even allowing for ''vanilla'' LO, all you are ever doing is the first step and so there is no need to explore multiple minima.

In summary, what I meant by "QPD assignments" was any **non-LO** QPD assignments.

circuit_knitting/cutting/cut_finding/best_first_search.py
```diff
@@ -149,6 +149,8 @@ class BestFirstSearch:
 
     ``stop_at_first_min`` (Boolean) is a flag that indicates whether or not to
     stop the search after the first minimum-cost goal state has been reached.
+    In the absence of any non-LO QPD assignments, it always makes sense to stop once
+    the first minimum has been reached and therefore, we set this bool to True.
```
garrison, line 153:
```suggestion
    the first minimum has been reached and therefore, we set this bool to ``True``.
```

circuit_knitting/cutting/cut_finding/best_first_search.py
```diff
@@ -268,7 +269,7 @@ def optimization_pass(
 
             self.update_minimum_reached(cost)
 
-            if cost is None or self.cost_bounds_exceeded(cost):
+            if cost is None or self.cost_bounds_exceeded(cost):  # pragma: no cover
```
garrison, line 272: Do you understand why this change resulted in this branch losing coverage?

ibrahim-shehzad: I think in exploring the multiple minima, it was encountering a ``None`` state for one of the test cases. I'd have to look into this more to figure out exactly what was going on.

garrison (approved):
Thanks, looks reasonable to me. :rocket:

caleb-johnson:
Could do a release note if we think the cut-finding performance is noticeable enough in general usage.

test/cutting/cut_finding/test_cut_finder_results.py
```diff
@@ -190,7 +190,7 @@ def test_four_qubit_circuit_two_qubit_qpu(
     )  # circuit separated into 2 subcircuits.
 
     assert (
-        optimization_pass.get_stats()["CutOptimization"] == array([15, 46, 15, 6])
+        optimization_pass.get_stats()["CutOptimization"] == array([11, 36, 15, 4])
```
caleb-johnson, line 193: Is there some sense I can make of these indices? (I think they're indices). How do you know what they should be here?

circuit_knitting/cutting/cut_finding/best_first_search.py
ibrahim-shehzad: If I remember correctly, when exploring multiple minima, the search ends when a `None` state is encountered. That doesn't happen when `stop_at_first_main` is set to `False`. I can check this early next week.

test/cutting/cut_finding/test_cut_finder_results.py
ibrahim-shehzad: They're basically just stats that keep track of things like how many states were added to the queue and how many backjumps were performed during the search. These numbers were obtained just by running the search algorithm on these circuits.

caleb-johnson: OK it's dubious to let the output of your function determine the target values in your unit tests. It's better to manually track down whether these values are correct when making the test case. I understand the temptation in situations like these, but doing this can give you a false sense of security and defeat the purpose of unit testing.

ibrahim-shehzad: I think it was partly also because I was not able to come up with a more creative test for this function.

caleb-johnson: I would probably remove the test if it were my code, since there could very well be a bug here that is treated as ground truth just because that was the output.

It's kind of outside the scope of this PR, but I just wanted to give my $.02 on that :D.

caleb-johnson (approved):
Awesome LGTM 🚀

circuit_knitting/cutting/cut_finding/best_first_search.py
ibrahim-shehzad: I have added a test for this case (here: [c4c2968](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/585/commits/c4c29685807ecf9f6709a2933c3c564446ecf2df)). The test also allows you to describe the workflow of the optimizer at a more granular level. We can force `BestFirstSearch.optimize()` to run until it encounters a `None` state which then tests this line. I am a little concerned, though, that I overdid this one a little bit and maybe excluding this line from coverage would have been okay instead.

caleb-johnson: I think it's a good test, thanks

test/cutting/cut_finding/test_cut_finder_results.py
```diff
@@ -189,9 +188,7 @@ def test_four_qubit_circuit_two_qubit_qpu(
         interface.export_subcircuits_as_string(name_mapping="default") == "AABB"
     )  # circuit separated into 2 subcircuits.
 
-    assert (
-        optimization_pass.get_stats()["CutOptimization"] == array([15, 46, 15, 6])
-    ).all()  # matches known stats.
+    assert optimization_pass.get_stats()["CutOptimization"][3] <= settings.max_backjumps
```
garrison, line 191: Why only consider the element at index `[3]`?

ibrahim-shehzad: It's really the only number (the number of backjumps) that is possible to (read easiest to) predict and constrain.

garrison: I see now: `get_stats` returns an array but every element of it means something specific.  This would have been better as a data structure or `namedtuple` rather than a numpy array.

ibrahim-shehzad: 'Tis done ([837875c](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/585/commits/837875c30bd056c202470426924b2dedb166d307)). Have also added a release note.

garrison: > 'Tis done ([837875c](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/585/commits/837875c30bd056c202470426924b2dedb166d307)). Have also added a release note.

Thank you.

We should probably actually revert this on this branch and then do this as a separate PR.  That way, we can backport this current PR since the interface does not change.  The improved interface can be done in 0.8.0 with a release note.

ibrahim-shehzad: Hmm this function isn't actually exposed in the API. Does that matter?

garrison: > Hmm this function isn't actually exposed through the API. Does that matter?

OK, in that case back-porting should be fine as is.

Sorry, I thought I saw something previously in the notebook where these same four numbers turned up, but that is either gone now or a phantom memory.

circuit_knitting/cutting/cut_finding/best_first_search.py
```diff
@@ -299,10 +314,10 @@ def minimum_reached(self) -> bool:
         """Return True if the optimization reached a global minimum."""
         return self.min_reached
 
-    def get_stats(self, penultimate: bool = False) -> np.typing.NDArray[np.int_] | None:
+    def get_stats(self, penultimate: bool = False) -> NamedTuple | None:
```
garrison, line 317: Does it complain if you use `SearchStats` as the return type?

circuit_knitting/cutting/cut_finding/best_first_search.py
```diff
@@ -312,15 +327,13 @@ def get_stats(self, penultimate: bool = False) -> np.typing.NDArray[np.int_] | N
...
-                self.num_backjumps,
-            ),
-            dtype=int,
+        search_stats = SearchStats(
+            states_visited=self.num_states_visited,
+            next_states_generated=self.num_next_states,
+            states_enqueued=self.num_enqueues,
+            backjumps=self.num_backjumps,
         )
+        return search_stats
```
garrison, line 336:
```suggestion
        return SearchStats(
            states_visited=self.num_states_visited,
            next_states_generated=self.num_next_states,
            states_enqueued=self.num_enqueues,
            backjumps=self.num_backjumps,
        )
```

circuit_knitting/cutting/cut_finding/best_first_search.py
ibrahim-shehzad: It doesn't, I have changed that here ([d931ec6](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/585/commits/d931ec60b33e730e290111c973552ed8c7111af2)).

circuit_knitting/cutting/cut_finding/lo_cuts_optimizer.py
```diff
@@ -155,10 +152,10 @@ def get_results(self) -> DisjointSubcircuitsState | None:
         """Return the optimization results."""
         return self.best_result
 
-    def get_stats(self, penultimate=False) -> dict[str, NDArray[np.int_]]:
+    def get_stats(self, penultimate=False) -> dict[str, NamedTuple | None]:
```
garrison, line 155: Is the idea that there might one day be more keys in this dict than just `CutOptimization`?

I wonder if this would be better.
```suggestion
    def get_stats(self, penultimate=False) -> dict[str, Any]:
```
but the docstring is still a little bit weird, because it talks about the "value" of the dict without referencing what the key(s) are.

ibrahim-shehzad: >Is the idea that there might one day be more keys in this dict than just CutOptimization?

Yes, that's right.

garrison: In that case, I think my suggestion of `Any` makes the most sense.

garrison (approved):
I am happy with this

PR #586, opened 2024-05-10
Provide options in the cut-finder API to turn LO gate and wire cut finding off or on, expose min-reached flag.

ibrahim-shehzad:
Add fields `gate_lo` and `wire_lo` to `OptimizationParameters` in `cutting.automated_cut_finding.py` to allow the user to turn LO gate and wire cutting off or on. The default is to have them both on.

To do:

- [x] Enabling only wire cutting causes multiple tests to fail. The same thing will likely happen if only gate cutting is enabled. Test both of these out and fix these tests.
- [x] Add tests for the `CutBothWires` action. While running the cut-finder on some circuits with `gate_lo` set to `False` (so only wire cut finding enabled), this action was implemented and we realized that there were some new bugs that were not covered by the existing tests.
- [x] Decide if pieces of the code that track the cost of LOCC wire cuts should be removed completely (cf https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/471#discussion_r1475017958).
- [x] Remove any pieces of the code that were excluded from coverage because they were tied to the rarely invoked `CutBothWires` instances.
- [x] Expose metadata field to check if minimum was reached, update tutorial, update tests, and clarify in release note that if the search is terminated early, this flag can give a false negative.


circuit_knitting/cutting/automated_cut_finding.py
```diff
@@ -137,6 +137,8 @@ class OptimizationParameters:
     seed: int | None = OptimizationSettings().seed
     max_gamma: float = OptimizationSettings().max_gamma
     max_backjumps: None | int = OptimizationSettings().max_backjumps
+    gate_lo: bool = OptimizationSettings().gate_lo
```
garrison, line 140: One might want to investigate what happens when one sets both `gate_lo` and `gate_lo` to `False`.  Ideally this will results in an informative error message (right?).

circuit_knitting/cutting/automated_cut_finding.py
```diff
@@ -137,6 +137,8 @@ class OptimizationParameters:
...
+    wire_lo: bool = OptimizationSettings().wire_lo
```
garrison, line 141: One might want to investigate what happens when one sets both `gate_lo` and `gate_lo` to `False`.  Ideally this will result in an informative error message (right?).

garrison:
>  ibrahim-shehzad added this to the 0.7.2 milestone

This adds a feature, so it (as a whole) will have to wait for 0.8.0.  It might be that some of the fixes are worth backporting, though.

circuit_knitting/cutting/automated_cut_finding.py
ibrahim-shehzad: Yes, for sure.

garrison: Also, acknowledging that we'll probably change the name of `gate_lo` eventually. since I doubt it makes any sense to implement LOCC gate cutting now that https://arxiv.org/abs/2312.11638 is out.

ibrahim-shehzad: I have gone ahead and added some of the updates that we discussed ([7fcfcb1](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/586/commits/7fcfcb1c051d9f6879709d9186f95323c9ebf4af)).

ibrahim-shehzad:
Instances of the greedy search (warm start) algorithm still had access to the LOCC wire-cut cost function and did, in certain instances, output costs that were inconsistent with the LO-only scenario that we currently support. This has now been changed to make things consistent and cost functions that track LOCC wire cuts are no longer exposed anywhere.

test/cutting/cut_finding/test_best_first_search.py
```diff
@@ -93,10 +93,11 @@ def test_best_first_search(test_circuit: SimpleGateList):
 
     out, _ = op.optimization_pass()
 
+    print(get_actions_list(out.actions))
```
garrison, line 96:
```suggestion
```

ibrahim-shehzad: Oh good catch :)

circuit_knitting/cutting/cut_finding/cut_optimization.py
```diff
@@ -60,11 +60,16 @@ def cut_optimization_cost_func(
...
+    goal_state: DisjointSubcircuitsState, func_args: CutOptimizationFuncArgs
 ) -> tuple[float, float]:
     """Return the value of :math:`gamma` computed assuming all LO cuts."""
     # pylint: disable=unused-argument
-    return (goal_state.upper_bound_gamma(), np.inf)
+    if goal_state is not None:
+        return (goal_state.upper_bound_gamma(), np.inf)
+    else:
+        raise ValueError(
+            "None state encountered: no cut state satisfying the specified constraints and settings could be found."
```
caleb-johnson, line 71: Very nitpicky, but can we capitalize "No"?

circuit_knitting/cutting/cut_finding/optimization_settings.py
```diff
@@ -38,16 +38,18 @@ class OptimizationSettings:
     If None is used as the random seed, then a seed is obtained using an
     operating-system call to achieve an unrepeatable randomized initialization.
 
-    NOTE: The current release only supports LO gate and wire cuts. LOCC
+    NOTE: The current release only supports LO gate and wire cuts. locc
```
caleb-johnson, line 41: Why did you go lower case here out of curiosity?

caleb-johnson (approved):
This looks good. Only a couple of very minor nitpicks.

I wonder if this could've been two PRs but looks good to me as-is

circuit_knitting/cutting/cut_finding/optimization_settings.py
ibrahim-shehzad: >Why did you go lower case here out of curiosity?

That was actually inadvertent, thank you for catching that. I went ahead and fixed that.

circuit_knitting/cutting/cut_finding/cut_optimization.py
```diff
@@ -125,7 +130,7 @@ def cut_optimization_goal_state_func(
 # Global variable that holds the search-space functions for generating
 # the cut optimization search space.
 cut_optimization_search_funcs = SearchFunctions(
-    cost_func=cut_optimization_cost_func,
+    cost_func=cut_optimization_upper_bound_cost_func,  # valid choice when considering only LO cuts.
```
garrison, line 133: How does this line changed relate to the change in #585?

ibrahim-shehzad: It's somewhat orthogonal but I think this is what (for consistency) it ought to have been from the start. It's hiding all LOCC related cost tracking.

garrison: Should this change be backported too?  What are the consequences of the stable release lacking this update?  (Are there any?)

ibrahim-shehzad: I don't think there are any consequences because the part that this effects was not exposed through our API. It's more just for the sake of consistency. I think we should be okay to not backport it.

garrison: Great; I thought that might be the case, just wanted to confirm :slightly_smiling_face:

ibrahim-shehzad: Always a good idea!

releasenotes/notes/min-reached-finder-flag-aa6dd9021e165f80.yaml
```diff
@@ -0,0 +1,10 @@
+---
+features:
+  - |
+    A new ``min_reached`` field has been added to the metadata outputted by :func:`circuit_knitting.cutting.find_cuts` to check if the cut-finder found
+    a cut scheme that minimized the sampling overhead. Note that the search algorithm employed by the cut-finder is `guaranteed` to find
```
garrison, line 5: Is you goal to italicize this?
```suggestion
    a cut scheme that minimized the sampling overhead. Note that the search algorithm employed by the cut-finder is *guaranteed* to find
```

releasenotes/notes/new-flags-to-control-cut-finder-search-e499e1ea49abb0bc.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    When specifying instances of :class:`~OptimizationParameters` that are inputted to :meth:`circuit_knitting.cutting.find_cuts()`, the user can now control whether the 
+    cut-finder looks only for gate cuts, only for wire cuts, or both, by setting the bools ``gate_lo`` and ``wire_lo`` appropriately. The default value
+    of both of these is set to `True` and so the default search considers the possibility of both gate and wire cuts.
```
garrison, line 6:
```suggestion
    of both of these is set to ``True`` and so the default search considers the possibility of both gate and wire cuts.
```

releasenotes/notes/min-reached-finder-flag-aa6dd9021e165f80.yaml
```diff
@@ -0,0 +1,10 @@
...
+    the optimal solution, that is, the solution with the minimum sampling overhead, provided it is allowed to run long enough.
+    The user is free to time-restrict the search by passing in suitable values for ``max_backjumps`` and/or ``max_gamma`` to
+    :class:`~OptimizationParameters`. If the search is terminated prematurely in this way, the metadata may indicate that the minimum
```
garrison, line 8: This reference doesn't currently resolve, but I expect this one to.
```suggestion
    :class:`.OptimizationParameters`. If the search is terminated prematurely in this way, the metadata may indicate that the minimum
```

releasenotes/notes/new-flags-to-control-cut-finder-search-e499e1ea49abb0bc.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    When specifying instances of :class:`~OptimizationParameters` that are inputted to :meth:`circuit_knitting.cutting.find_cuts()`, the user can now control whether the
```
garrison, line 4: Likewise
```suggestion
    When specifying instances of :class:`.OptimizationParameters` that are inputted to :meth:`circuit_knitting.cutting.find_cuts()`, the user can now control whether the 
```

circuit_knitting/cutting/automated_cut_finding.py
```diff
@@ -126,6 +132,7 @@ def find_cuts(
         elif inst.operation.name == "cut_wire":
             metadata["cuts"].append(("Wire Cut", i))
     metadata["sampling_overhead"] = opt_out.upper_bound_gamma() ** 2
+    metadata["min_reached"] = optimizer.minimum_reached()
```
garrison, line 135: I wonder if we should just call it `minimum_reached` to avoid an abbreviation.

```suggestion
    metadata["min_reached"] = optimizer.minimum_reached()
```

garrison: Yes (and also the string needs to be modified in I believe 5 other locations).

test/cutting/test_find_cuts.py
```diff
@@ -46,6 +47,24 @@ def test_find_cuts(self):
             assert len(metadata["cuts"]) == 2
             assert {"Wire Cut", "Gate Cut"} == cut_types
             assert np.isclose(127.06026169, metadata["sampling_overhead"], atol=1e-8)
+            assert metadata["min_reached"]
```
garrison, line 50: This is a stronger condition, as the way you previously wrote it would pass if `metadata["min_reached"]` is any ["truthy" value](https://docs.python.org/3/library/stdtypes.html#truth-value-testing).

```suggestion
            assert metadata["min_reached"] is True
```

ibrahim-shehzad: I'll double check but I think the reason why I didn't do this earlier was because I was getting a linting error.

ibrahim-shehzad: Nevermind, I think I was hallucinating there. I have made the changes discussed here ([53deb86](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/586/commits/53deb866f797d69868e59ec87091747af5f2a309)).

circuit_knitting/cutting/automated_cut_finding.py
ibrahim-shehzad: I have gone ahead and done this.

test/cutting/test_find_cuts.py
garrison: > I was getting a linting error.

I wouldn't be surprised if you saw this elsewhere when using something like this.  In general, it's much better to e.g. condition a code block on truthiness rather than `is True`; but this is one of those special cases (in a test) where it actually makes sense to make sure that it `is True`.

circuit_knitting/cutting/cut_finding/disjoint_subcircuits_state.py
```diff
@@ -231,7 +233,7 @@ def cut_actions_sublist(self) -> list[NamedTuple]:
                 self.cut_actions_list.append(
                     CutIdentifier(
                         cut_actions[i].action.get_name(),
-                        GateCutLocation(
+                        CutLocation(
```
garrison, line 236: I think the comment on line 232 could be updated.

garrison: OK, that is reasonable.

garrison (approved):
Thank you -- I am happy with this. :rocket:

PR #589, opened 2024-05-15
Remove use of qasm simulator

### LINKED ISSUE #502: Remove uses of `ibmq_qasm_simulator` from cutqc tutorials
Cloud simulators are being retired: https://docs.quantum.ibm.com/announcements/product-updates/2024-03-22-announcement-cloud-simulators-and-lab

garrison:
Closes #502.

I had not noticed before, but the section on setting up the sampler primitive appears twice in the second cutqc tutorial.  So I was able to remove the first appearance.


circuit_knitting/cutting/cutqc/wire_cutting_evaluation.py
```diff
@@ -81,7 +81,8 @@ def run_subcircuit_instances(
                     options[i % len(options)] for i, _ in enumerate(subcircuits)
                 ]
         else:
-            backend_names_repeated = ["ibmq_qasm_simulator"] * len(subcircuits)
+            least_busy = service.least_busy().name
+            backend_names_repeated = [least_busy] * len(subcircuits)
```
garrison, line 85: Actually, I am going to change this to restore the old code path and just put a warning right before it.  I am doing this with the typical persona of somebody using this (deprecated) code in mind.  If I were using cutqc still, I'd be very surprised if I upgraded and my old code that used a cloud qasm simulator suddenly uses a hardware backend instead.  Better is for the user to have an error, occur, and be able to deal with it as they see fit.  With the warning, they'll have a bit more context for why the qasm simulator is suddenly failing and what to do to fix it (pass `backend_names`).

garrison: Updated accordingly in 2d27024daede52a7cb3664e5db973b6dde981e46.

caleb-johnson (approved):
Thanks!

PR #591, opened 2024-05-15
Tuple unpacking bug fix in cut finder.

ibrahim-shehzad:
There was a tuple unpacking error that arose when the`CutBothWires` action was invoked as part of the rest of the cut finding workflow. This bug was not noticed at first because the although existing tests called that line, its output was not used in any part of the rest of the workflow. This bug was noticed in the work leading up to #586  which will also add more comprehensive tests for the `CutBothWires` action. The goal of this PR is just to fix this bug and edit the corresponding tests and type hints accordingly.

### DIFF
--- a/circuit_knitting/cutting/cut_finding/disjoint_subcircuits_state.py
+++ b/circuit_knitting/cutting/cut_finding/disjoint_subcircuits_state.py
@@ -37,7 +37,7 @@ class Action(NamedTuple):
 
     action: DisjointSearchAction
     gate_spec: GateSpec
-    args: list
+    args: list | tuple
 
 
 class GateCutLocation(NamedTuple):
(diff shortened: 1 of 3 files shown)


circuit_knitting/cutting/cut_finding/disjoint_subcircuits_state.py
```diff
@@ -430,12 +430,12 @@ def add_action(
         self,
         action_obj: DisjointSearchAction,
         gate_spec: GateSpec,
-        args: tuple | None = None,
+        *args: tuple | None,
```
garrison, line 433: I suspect that `| None` is not actually needed.  See e.g. https://stackoverflow.com/a/37032111/1558890

caleb-johnson (approved):
LGTM, thanks

circuit_knitting/cutting/cut_finding/disjoint_subcircuits_state.py
ibrahim-shehzad: I wasn't able to look into this careful but since `mypy` didn't complain, I went ahead and made the change you suggested ([e44b4be](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/591/commits/e44b4bedb6f228347da5e602e57c24c9dee0159d)).

garrison (approved):

PR #594, opened 2024-05-18
Remove idle wires by default when separating the circuit

garrison:
Currently, if one has a 10-qubit circuit and only 2 qubits are used, then `partition_problem` will separate the input into 8 circuits that have a single unused qubit, plus one or two circuits (depending on connectivity) for the 2 qubits actually used.  This changes the default to remove idle wires.  It also means that the partition label of `None` will be treated specially (it can only be used on an idle wire).

#### Remaining action items

- [x] Elaborate on release note


circuit_knitting/utils/transforms.py
```diff
@@ -158,12 +167,26 @@ def _partition_labels_from_circuit(
...
+        idle_wires = set(range(circuit.num_qubits))
+        for instruction in circuit.data:
+            for q1 in instruction.qubits:
+                q1_id = circuit.find_bit(q1).index
+                idle_wires.discard(q1_id)
+        # Replace qubit_subsets with filtered list, removing idle qubits
+        qubit_subsets = [
+            subset
+            for subset in qubit_subsets
+            if not (len(subset) == 1 and next(iter(subset)) in idle_wires)
```
garrison, line 182: We have to use `next(iter(subset))` here because that is the standard way to get the "first" element from a `set`.

circuit_knitting/utils/transforms.py
```diff
@@ -51,11 +51,12 @@ class SeparatedCircuits(NamedTuple):
...
     Each element of that list is a 2-tuple which includes the partition label
     of that qubit, together with the index of that qubit in the corresponding
-    subcircuit.
+    subcircuit.  If the original qubit is unused and has been removed from the separated circuits,
+    then that tuple will be equal to ``(None, None)``.
     """
 
     subcircuits: dict[Hashable, QuantumCircuit]
-    qubit_map: list[tuple[Hashable, int]]
+    qubit_map: list[tuple[Hashable, int] | tuple[None, None]]
```
garrison, line 59: It turns out to be a lot easier to work with this if it's always a 2-tuple (which means you can always unpack it), which is which I've used `(None, None)` instead of `None`.

circuit_knitting/utils/transforms.py
```diff
@@ -191,44 +214,60 @@ def _circuit_from_instructions(
...
-        }
-        # All qubits for any gate should belong to one subset
+    for i, inst in enumerate(circuit.data):
+        # Collect the partition labels spanned by the instruction
+        partitions_spanned = set()
+        for qubit in inst.qubits:
+            j = circuit.find_bit(qubit).index
+            partition_id = qubit_map[j][0]
+            if partition_id is None:
+                raise ValueError(
```
caleb-johnson, line 248: Do you raise here because `None` is technically `Hashable`? Users could pass in a `Sequence[None]`, and it would obey the `Sequence[Hashable]` type hint, but be an invalid input?

garrison: > Do you raise here because `None` is technically `Hashable`? Users could pass in a `Sequence[None]`, and it would obey the `Sequence[Hashable]` type hint, but be an invalid input?

Not quite; it's because a label of `None` means "this qubit should be idle," but if this code path is triggered, it's because there is an instruction that uses the qubit in the circuit (i.e., it's not _actually_ idle).

I wonder if the error message (on the following lines) could be clarified in any way.

caleb-johnson: No, I think it's a good message. This arises when a user passes `None` as a label for an active qubit, and I think that's clear here.

I guess what I was getting at is that the user passing `None` on an active qubit accidentally doesn't disobey the type hint, so raising a `ValueError` is appropriate.

caleb-johnson (approved):
This looks nice, thanks! 🚀

PR #597, opened 2024-05-22
Fix coverage workflow and remove workflow dependency on coveragepy-lcov

garrison:
It looks like we no longer even need to coveragepy-lcov tool: https://github.com/nedbat/coveragepy/issues/587#issuecomment-1019370754

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -34,10 +34,10 @@ jobs:
         run: |
           tox -e coverage
       - name: Convert to lcov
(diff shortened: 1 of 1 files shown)


garrison:
I am beyond puzzled about why the coveralls upload suddenly stopped working.  I found https://github.com/coverallsapp/github-action/issues/207#issuecomment-2087527849 but it looks like we are doing everything by the book.

garrison:
I suspect the problem is https://github.com/coverallsapp/coverage-reporter/issues/127.

PR #598, opened 2024-05-22
Tutorials: Fix typo and split cell that retrieves results

garrison:
This fixes a typo and splits the cell so that the results are retrieved in their own cell.  This is nice because then the user will have a cell complete once all jobs are submitted and another complete when all results are available.  I also removed an outdated reference to the old "commented block."

### DIFF
--- a/docs/circuit_cutting/tutorials/01_gate_cutting_to_reduce_circuit_width.ipynb
+++ b/docs/circuit_cutting/tutorials/01_gate_cutting_to_reduce_circuit_width.ipynb
@@ -314,14 +314,14 @@
    "source": [
     "from qiskit_ibm_runtime import SamplerV2, Batch\n",
     "\n",
-    "# Set up a Qiskit Runtime Sampler primitive for each circuit partition\n",
-    "samplers = {label: SamplerV2(backend=backend) for label in subexperiments.keys()}\n",
+    "# Set up the Qiskit Runtime Sampler primitive\n",
+    "sampler = SamplerV2(backend=backend)\n",
     "\n",
     "# Submit each partition's subexperiments as a single batch\n",
     "with Batch(backend=backend):\n",
     "    jobs = {\n",
-    "        label: sampler.run(isa_subexperiments[label], shots=2**12)\n",
-    "        for label, sampler in samplers.items()\n",
+    "        label: sampler.run(subsystem_subexpts, shots=2**12)\n",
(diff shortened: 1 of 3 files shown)


docs/circuit_cutting/tutorials/01_gate_cutting_to_reduce_circuit_width.ipynb
```diff
@@ -322,9 +322,17 @@
     "    jobs = {\n",
```
ibrahim-shehzad, line 322: This probably a little unrelated to the scope of this PR but I wanted to point this out anyway. I encountered this while running something in batch mode recently. 
1. Should `samplers` be defined inside the context manager specified by `Batch`? I noticed that when I did:

```
estimators = {edge: Estimator(backend=backend) for edge in list_of_edges}
with Batch(backend=backend): 
     jobs = {edge: estimator.run([(xpiled_circuit, experiments[edge])])
        for edge, estimator in estimators.items()}
```
my jobs did not get a batch id. 

    
   2. Secondly, do we need independent instantiations of `SamplerV2` for each subexperiment?

garrison: > do we need independent instantiations of SamplerV2 for each subexperiment?

No, we don't.  This is legacy from when we wanted to give the user an option to run across two different hardware devices.  But this is not compatible with `Batch` mode anyway.

If we have a single sampler created outside the context manager, does this fix point 1?

ibrahim-shehzad: >If we have a single sampler created outside the context manager, does this fix point 1?

I would have to test this out but what I suggested above was recommended to me by a colleague (Saki).

garrison: I got rid of the separate samplers.  I am still creating the sampler outside of the context manager because `batch.session.id` seems to get set either way.  If we find that there is an additional improvement to be made here, we can always make a follow up PR.

ibrahim-shehzad: Okay, that makes sense.

ibrahim-shehzad (approved):
This looks good to me, thanks.

PR #611, opened 2024-05-30
Add note in INSTALL regarding Docker installation

caleb-johnson:
This note is at least useful for developers who will commonly have large `.tox` directories in their local repository, along with other caches which speed up local CI.

This could also be useful for someone running a local install and using the repository as a dumping grounds for results.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -100,6 +100,11 @@ build a Docker image, as well as a
 `compose.yaml <https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/main/compose.yaml>`__ file, which allows one
 to use the Docker image with just a few simple commands.
 
+.. note::
+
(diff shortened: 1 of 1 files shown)


garrison:
> This note is at least useful for developers who will commonly have large `.tox` directories in their local repository, along with other caches which speed up local CI.

Are you seeing this with `.tox` specifically?  `.tox/` is listed in [.dockerignore](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/main/.dockerignore), but maybe we need to list `**/.tox` instead.

garrison:
(I am not opposed to adding a note, but I do want to fix the problem to the extent possible that we can.)

caleb-johnson:
No, you are right. This came up developing in another repo. We are covered here. Closing

PR #612, opened 2024-05-30
Remove `basepython` restriction in `lint` and `coverage` environments

garrison:
This restriction was previously necessary, but it no longer appears to be, as far as I can tell.

Getting rid of the restriction will make these environments work on systems where `python3.10` is not installed.


caleb-johnson (approved):
Cool LGTM

PR #613, opened 2024-05-31
Move ruff and mypy configuration to `pyproject.toml`

garrison:
We already use `pyproject.toml` to configure autoflake, pylint, coverage, and pytest.  This moves ruff and mypy configuration to be in `pyproject.toml` as well, to centralize the linter configurations.

If we're going to be maintaining multiple repositories, it will be easiest to compare linting standards if the tool configurations are all specified in a single file in each repository.


caleb-johnson (approved):
Agreed, thanks! I wasn't sure if there were some options which misbehaved from pyproject.toml. It seems like I have encountered that a time or two. Seems like we're safe to switch though 👍

PR #616, opened 2024-06-01
Use batch mode correctly in the tutorials

garrison:
Apparently, if you use `sampler = SamplerV2(backend=backend)`, even within the `Batch` context manager, Qiskit Runtime will use job mode instead of batch mode, without any warning.  This fixes our example usage so batch mode will work properly.


garrison:
I tested with `QiskitRuntimeService`, and a session was indeed created.  But CI fails because it no longer works on a fake backend, in local mode.

ibrahim-shehzad:
>even within the Batch context manager, Qiskit Runtime will use batch mode instead of job mode, without any warning. 

Did you mean Qiskit Runtime will use _job_ mode instead of _batch_ mode when this is the case? This was what I was running into.

garrison:
> > even within the Batch context manager, Qiskit Runtime will use batch mode instead of job mode, without any warning.
> 
> Did you mean Qiskit Runtime will use _job_ mode instead of _batch_ mode when this is the case? This was what I was running into.

I did.  I fixed it in the top comment.

garrison:
> I'm guessing we will change to
> 
> ```
> with Batch(backend=backend) as batch: 
>     sampler = SamplerV2(mode=batch)
> ```
> 
> after runtime makes that change and backport that?

It probably won't be a candidate for backporting, as I expect it to require us to bump the minimum required version of `qiskit-ibm-runtime`.

PR #623, opened 2024-06-12
Bump qiskit-ibm-runtime version and use `mode=batch`

garrison:
This is necessary to prevent a `DeprecationWarning` under qiskit-ibm-runtime 0.24, which is currently visible in the [first](https://qiskit-extensions.github.io/circuit-knitting-toolbox/circuit_cutting/tutorials/01_gate_cutting_to_reduce_circuit_width.html) and third tutorials.  The warning is a result of the merging of https://github.com/Qiskit/qiskit-ibm-runtime/pull/1584.

qiskit-ibm-runtime 0.24 depends on qiskit>=1.1, so I had to bump the qiskit version as well.

Technically, this version bump is only required by the notebooks, not the actual CKT library, but the "minimum version tests" aren't equipped to deal with situations where the notebooks have a different minimum version than the CKT library.


ibrahim-shehzad:
This looks great but I am wondering if it may be helpful to say at the start, what exactly the `DeprecationWarning` was. I am guessing part of it was related to the use of `mode = batch` in place of `session = batch`.

ibrahim-shehzad (approved):
This looks good to me, thanks!

PR #625, opened 2024-06-14
transforms.py: Avoid setting the barrier to be a uuid object

garrison:
More info at https://github.com/Qiskit/qiskit/issues/12581


garrison:
I'm going to merge this immediately so CI passes once again.

PR #605, opened 2024-05-22
Remove cutQC

caleb-johnson:
Fixes #519 

#### Remaining action items

- [x] Release note
- [x] Remove `utils/conversion.py` and `utils/metrics.py`
- [x] Remove `--omit="circuit_knitting/cutting/cutqc/**/*,circuit_knitting/utils/conversion.py,circuit_knitting/utils/metrics.py"` from coverage command in `tox.ini`
- [x] Remove `cutqc` and `metrics` from `addopts` line in `pyproject.toml`, and anything else added temporarily as part of #527
- [x] grep for remaining references to cplex (e.g. in the README)

PR #631, opened 2024-07-08
Pin numpy<2 on stable branch

garrison:
From the looks of the CI failure, DOCplex is incompatible with numpy 2.0.  So, as a way to make the CutQC code continue to work, I am pinning to `numpy <2` on the `stable/0.7` branch.  (CutQC has already been removed from `main`, in #605.)


garrison:
I'm just going to merge this, because I see no alternative (other than allowing CutQC and CI to be broken).

Issue #621, opened 2024-06-07
An idle qubit is still idle even if it has a barrier


garrison:
I noticed in the fragment of code for identifying idle qubits that it will identify a qubit as non-idle if it has a barrier, even if nothing else is on it:

https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/blob/67f24cb4e8e067eebad1f351ce971df8dc115f61/circuit_knitting/utils/transforms.py#L170-L177

This is in contrast to the proposed similar method in https://github.com/Qiskit/qiskit/pull/12470.

Issue #626, opened 2024-06-15
Need test that barriers and re-combined correctly in `transforms.py`


garrison:
See https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/625#issuecomment-2168475767

lodenrogue:
Any more details you can provide here? An example of what is meant by a barrier or re-combined for those of us new to this repo?

garrison:
If you take the first tutorial notebook and place a barrier across all qubits at some point in the circuit, there is some special logic to split that barrier into single-qubit barriers so that it will not prevent the circuit from separating into multiple pieces. Then, after separation, those barriers are recombined. #625 fixed this code to work with Qiskit 1.2.  It is the stuff in `transforms.py` that has to do with uuids.

garrison:
The test should ideally verify the behavior of the `seprate_circuit` utility function, which is called inside of the `partition_problem` function, which itself appears in the first tutorial. And the test should live in `teat_transforms.py`, alongside other tests of the `separate_circuit` function.

garrison:
The crux of the problem is that the test suite should fail if the following patch is made to disable (i.e., comment out) an important line, but it does not.

```diff
diff --git a/circuit_knitting/utils/transforms.py b/circuit_knitting/utils/transforms.py
index 2849d09..715321a 100644
--- a/circuit_knitting/utils/transforms.py
+++ b/circuit_knitting/utils/transforms.py
@@ -135,7 +135,7 @@ def separate_circuit(
             [new_qc.qubits[j] for j in qubits_by_subsystem[label]],
             new_qc.cregs,
         )
-        _combine_barriers(tmp_circ)
+        # _combine_barriers(tmp_circ)
         subcircuits[label] = tmp_circ
 
     return SeparatedCircuits(subcircuits, qubit_map)
```

PR #639, opened 2024-07-16
Fix idle qubit detection to ignore barriers

raunakkumarsingh:
Check for Barriers:
Added if instruction.operation.name == 'barrier': continue to skip over barrier operations during the idle wire detection process.
Iterate Through Instructions:

For each instruction, if it is not a barrier, it iterates through the qubits involved in the instruction and removes those qubit indices from the idle_wires set.
Filter Qubit Subsets:

The qubit_subsets list is filtered to remove any subsets that consist of only one qubit, and that qubit is in the idle_wires set.


raunakkumarsingh:
Added if instruction.operation.name == 'barrier': continue to skip over barrier operations during the idle wire detection process.

garrison:
It'd be nice to have an explicit test for this behavior, too.

garrison:
The lint failure can be fixed by merging `main` into this branch once #640 has been merged to `main`.

garrison:
@raunakkumarsingh, thank you for this contribution.  Do you plan to add a test and merge `main` into this branch, or should I do it?

raunakkumarsingh:
I have added tests for the barrier Ignore and rebased my branch.

garrison:
Thank you.  Please run `tox -estyle` and make sure that `tox -elint` passes.

raunakkumarsingh:
corrected lint

test/utils/test_transforms.py
```diff
@@ -262,6 +262,48 @@ def test_separate_circuit(self):
...
+            circuit = QuantumCircuit(qreg)
+            circuit.h(0)
+            circuit.x(0)
+            circuit.barrier(0)
+            circuit.x(1)
+            circuit.h(1)
+            circuit.y(2)
+            circuit.h(2)
+
+            separated_circuits = separate_circuit(circuit, partition_labels="ABC")
```
garrison, line 276: Actually, I believe this test does not actually check that the relevant bug is fixed.

The idle wire detection code only runs when `partition_labels` is not provided (or is `None`).  In this case, the partition labels are determined automatically from the connectivity of the circuit.   #594 improved on this code so that idle wires are removed from the circuit during separation, but the problem is that any qubit with a barrier (e.g., if a barrier is placed over all qubits, even the idle ones) will be falsely detected as an active qubit.  Your change should indeed fix #621 but the test does not ensure that it is fixed.  The test is actually _almost_ testing the check desired in #626, but not quite.

Issue #506, opened 2024-03-15
Use SamplerV2 and EstimatorV2 in tests


garrison:
Eventually -- once support is added for `SamplerV2` (#488) and once Qiskit Aer supports `SamplerV2` and `EstimatorV2` (https://github.com/Qiskit/qiskit-aer/issues/2078) -- we should update the tests to use the v2 versions of the primitives.

At the moment, I expect the migration to `EstimatorV2` to be relatively easy.

`SamplerV2`, however, is going to be a bit more involved.  `ExactSampler` returns exact probabilities, but `SamplerV2` has moved away from probabilities and instead provides memory-based access to each shot outcome.  So, the idea of an "exact sampler" is incompatible with the v2 sampler interface.  Thus, we will probably stick with `ExactSampler` using the V1 interface for as long as we can.  Once the V1 sampler interface is deprecated, we will be forced to upgrade, and this might mean we are no longer able to use the Sampler interface for the exact tests from this point forward.  This is probably not a big deal, but it will be a minor inconvenience for us.  Still, the v2 sampler is unambiguously a step forward, so I eagerly await the day when we can use it exclusively.

garrison:
Actually, since `Sampler` (V1) is in Qiskit 1.0, it won't be removed until Qiskit 2.0 at the earliest.  So we have a year before we might actually be forced to do something here.

ihincks:
I'd like to understand this issue better. Are there some examples you might recommend?

garrison:
> I'd like to understand this issue better. Are there some examples you might recommend?

We have a class called [`ExactSampler`](https://qiskit-extensions.github.io/circuit-knitting-toolbox/stubs/circuit_knitting.utils.simulation.ExactSampler.html), which acts just like `qiskit.primitives.Sampler` (v1) does when `shots=None`, except it has workarounds for a few limitations in `qiskit.primitives.Sampler`.  It provides exact probabilities, which is useful for us in two circumstances:

[...]

PR #643, opened 2024-07-17
Use Aer's SamplerV2 instead of SamplerV1

garrison:
Addresses part of #506.

#### Remaining action items

- [x] get it to work
- [ ] possibly also update an EstimatorV1 use in the roundtrip tests.


garrison:
This is pretty obviously a step forward; I am just going to merge it.

PR #683, opened 2024-09-12
fix: broken link in Release shield

mrossinek:
There are also more mentions of `Qiskit-Extensions/circuit-knitting-toolbox` in some old release notes but I did not address those yet.

Those links appear to be direct github links to jupyter notebooks rather than links pointing to their HTML rendered versions. Might be a bit more involved to address.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 <!-- SHIELDS -->
 <div align="left">
 
-  [![Release](https://img.shields.io/pypi/v/qiskit-addon-cutting.svg?label=Release)](https://github.com/Qiskit-Extensions/qiskit-addon-cutting/releases)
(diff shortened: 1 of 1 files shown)


garrison:
Thanks, I don't know how I missed this one.

> There are also more mentions of Qiskit-Extensions/circuit-knitting-toolbox in some old release notes but I did not address those yet.
>
> Those links appear to be direct github links to jupyter notebooks rather than links pointing to their HTML rendered versions. Might be a bit more involved to address.

I did not make an effort to update names and external links in old release notes.  However, I did make an effort to ensure that redirects are in place, at least from the pages that were in place at the time of the project was renamed: https://github.com/Qiskit-Extensions/qiskit-extensions.github.io/pull/2.

garrison (approved):

PR #693, opened 2024-10-23
Prepare API docs links for ibm.com migration

Eric-Arellano:
Relative links to non-API docs will break when migrating to docs.quantum.ibm.com. They should instead be absolute links.

There are also two links to docs.quantum.ibm.com/api/migration-guides. That was migrated from `/api/migration-guides` to `/migration-guides`. While there's a redirect, we should use the new URL directly.


garrison (approved):
Thanks!  LGTM

PR #702, opened 2024-10-30
Show class members

Eric-Arellano:
We weren't showing class members (attributes & methods) for inlined classes. This is fixed by setting `"inherited-members": None` in `autodoc_default_options`. 

Notably, we now show inherited members for classes like `instructions.Move`. I recommend this because users want to know every function/attribute defined on the object, and they don't really care about the origin of that functionality. Qiskit SDK takes this approach, which is useful with e.g. the circuit library.

With this change, some of the classes end up being much larger. So, we un-inline them and move them back to being their own pages in `stubs/`. This means we remove their client redirects.

This PR also makes some other enhancements:

* Adds `tox -e docs-clean`
* Displays type hints in the `parameters` section, rather than inline to the signature. This matches the style of the other API projects


docs/_templates/autosummary/class.rst
```diff
@@ -0,0 +1,33 @@
+{#
```
Eric-Arellano, line 1: This is copied from Qiskit SDK, other than where I comment. I'm probably going to update Qiskit to mirror this

docs/_templates/autosummary/class.rst
```diff
@@ -0,0 +1,33 @@
...
+-#}
+
+{{ objname | escape | underline }}
+
+.. currentmodule:: {{ module }}
+
+.. autoclass:: {{ objname }}
+   :no-members:
+   :no-inherited-members:
+   :no-special-members:
```
Eric-Arellano, line 13: We have to set this to avoid the methods being duplicated. That's because we configured `.. autoclass` in `conf.py` to now show inherited members

docs/_templates/autosummary/class_no_inherited_members.rst
Eric-Arellano, line 1: This was already unused

docs/apidocs/qiskit_addon_cutting.qpd.rst
```diff
@@ -9,12 +9,16 @@ Quasi-Probability Decomposition (QPD) (:mod:`qiskit_addon_cutting.qpd`)
...
+.. autosummary::
+   :nosignatures:
+   :toctree: ../stubs/
+
+   QPDBasis
+   BaseQPDGate
+   SingleQubitQPDGate
+   TwoQubitQPDGate
 
+.. autoclass:: WeightType
```
Eric-Arellano, line 21: This class is tiny, so I think it's fine to keep inlined

docs/conf.py
```diff
@@ -83,10 +83,16 @@
...
+# Options for autodoc. These reflect the values from Qiskit SDK and Runtime.
 autosummary_generate = True
 autosummary_generate_overwrite = False
 autoclass_content = "both"
+autodoc_typehints = "description"
+autodoc_default_options = {
+    "inherited-members": None,
+}
+napoleon_google_docstring = True
+napoleon_numpy_docstring = False
```
Eric-Arellano, line 95: This is a performance enhancement for build-time

qiskit_addon_cutting/utils/observable_grouping.py
```diff
@@ -250,13 +250,13 @@ def groups(self) -> list[CommutingObservableGroup]:
 
     @property
     def lookup(self) -> dict[Pauli, list[tuple[int, int]]]:
-        r"""Get dict which maps each :class:`~qiskit.quantum_info.Pauli` observable to a list of indices, ``(i, j)``, to commuting observables in ``groups``.
+        """Get dict which maps each :class:`~qiskit.quantum_info.Pauli` observable to a list of indices, ``(i, j)``, to commuting observables in ``groups``.
 
         For each element of the list, it means that the :class:`~qiskit.quantum_info.Pauli` is given by
-        the ``j``th commuting observable in the ``i``th group.
+        the ``j``-th commuting observable in the ``i``-th group.
```
Eric-Arellano, line 256: The lack of `-` was causing a Sphinx error

qiskit_addon_cutting/utils/observable_grouping.py
```diff
@@ -250,13 +250,13 @@ def groups(self) -> list[CommutingObservableGroup]:
...
 
         This list will be of length 1 at minimum, but may potentially be longer
-        if multiple :class:`.CommutingObservableGroup`\ s are compatible with the given
+        if multiple :class:`.CommutingObservableGroup` objects are compatible with the given
```
Eric-Arellano, line 259: Best practice is to pluralize like this

garrison: I don't agree with this change.  Can you provide some more justification than "best practice"?

If I read it aloud, the phrase "multiple commuting observable groups are compatible with the given Pauli" makes sense to me as a physicist (other than Qiskit's weird convention of using "group" rather than "set").  I am not sure what value "objects" brings.

Eric-Arellano: The `\s` style with code before it is hard to read and especially hard to globalize/translate. Abby C and Becky were concerned about it. So, the alternative is to use the format "blah blah `MyClass` objects/intances"

> I am not sure what value "objects" brings.

Here, objects == instances of the class. Is there an alternative word you prefer like "instances"?

> hard to globalize/translate

Fwiw, we haven't finalized yet whether we will globalize API docs. Still, it's preferable to use a consistent style between API docs and guides

This isn't a hard requirement - you can push back

tox.ini
```diff
@@ -59,3 +59,10 @@ commands =
   sphinx-build -j auto -W -T --keep-going {posargs} docs/ docs/_build/html
 passenv =
   CI
+
+[testenv:docs-clean]
+skip_install = true
+allowlist_externals =
+  rm
+commands =
+  rm -rf {toxinidir}/docs/stubs/ {toxinidir}/docs/_build/
```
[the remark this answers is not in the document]
Eric-Arellano: Ack, I noticed that. It only deletes `.doctrees` inside `html` rather than the entire folder, however. I also think it's nice to have `tox -e docs-clean` to be consistent with the other addon repos - muscle memory.

Lmk if you want me to delete this though. I'm fine with that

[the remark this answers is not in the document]
garrison: Sorry, I meant that we could potentially remove those steps from the `docs` environment now that there is docs-clean.

garrison (dismissed):
This looks good to me

garrison (approved):

PR #705, opened 2024-11-04
Update some Qiskit Terra links to Qiskit repo

Eric-Arellano:
The qiskit/documentation external link checker did not like a link to qiskit-terra, which is now qiskit. While it redirects fine, we intentionally try to avoid linking to redirects.


Eric-Arellano (commented):
I don't know why the link checker picked up on it this week and not the week before 🤷

garrison (approved):

PR #743, opened 2025-03-07
Fix deprecations for Qiskit 2.0

garrison:


test/cut_finding/test_cut_finder_results.py
garrison, line 1: @ibrahim-shehzad, the change to `efficient_su2` means the gates are in a slightly different order -- see the changes to `test/cut_finding/test_cco_utils.py`.  With that in mind, do these changes to the tests in this file look reasonable to you?

ibrahim-shehzad: In these tests, I obtained `instruction_id's` by running the cut-finder on the circuits used in these tests, looking at the output and checking if the result made sense. White I have not repeated that based on these most recent changes, I do think it looks reasonable. Let me know if there is a more specific check you would like me to do.

garrison: Thanks, I think it looks reasonable to me too.

PR #775, opened 2025-09-24
Undo the Qiskit 2.2 workaround added in #773

garrison:
This reverts to parallel pass managers, but CI is expected to fail until something else is fixed, either in this package or in Qiskit.


garrison:
Qiskit 2.2.1 has been released, so this can be merged as soon as CI turns green.

PR #818, opened 2026-06-11
Ajc/updates to docs for migration

abbycross:
Preparing addon docs for migration to IBM Quantum Platform


docs/explanation/index.rst
```diff
@@ -130,17 +130,17 @@ Wire cutting phrased as a two-qubit :class:`.Move` operation
...
+Much of the circuit cutting literature describes a process where we sample from the distribution and take a single shot, then sample from the distribution again and repeat; however, this is not feasible in practice, so we instead perform all sampling upfront.  For now, because of limitations in version 1 of the Qiskit primitives, we take a fixed number of shots for each considered subexperiment and send the subexperiments to the backend(s) in batches. During reconstruction, each subexperiment contributes to the final result with proportion equal to its weight.  We must ensure the number of shots taken is sufficient for the heaviest weighted subexperiment.  In the future, we plan to support passing an individual ``shots`` count with each subexperiment to Qiskit Runtime, so that each subexperiment will be run with a number of shots proportional to that subexperiment's weight in the final result (see issue `#532 <https://github.com/Qiskit/qiskit-addon-cutting/issues/532>`__).  This per-experiment shots count is a new feature enabled by version 2 of the Qiskit primitives.
```
abbycross: Does this require any updating in:re primitives, or is this all still true?

INSTALL.rst
```diff
@@ -1,21 +1,19 @@
...
 - :ref:`Option 1`
 - :ref:`Option 2`
 - :ref:`Option 3`
 
-Users who wish to run within a containerized environment may skip the
+If you want to run within a containerized environment, you can skip the
 pre-installation and move straight to :ref:`Option 3`.
 
-Pre-Installation
+Pre-installation
```
abbycross:
```suggestion
Prerequisites
```

abbycross:
Ready for review @garrison @jenglick

docs/explanation/index.rst
garrison: The primitives now support this, but this addon still isn't able to leverage that feature -- that's what the issue is about.  So I think the wording should be updated to reflect that the primitives v2 have the necessary feature, and there is an open issue to allow the addon to leverage this.  I suggest removing the phrase "In the future, we plan to support" because it is not clear that anyone has plans at this point to fix this issue.

By the way, I expect that the `&mdash;`s will not render properly in Sphinx/restructedtext.  Would you be open to using either three contiguous hyphens (`---`) or a unicode mdash?

docs/explanation/index.rst
```diff
@@ -6,29 +6,29 @@ Explanatory material
 
 Overview of circuit cutting
 ---------------------------
-Circuit cutting is a technique to increase the size of circuits we can run on quantum hardware at the cost of an additional sampling overhead. A larger quantum circuit can be decomposed by cutting its gates and/or wires, resulting in smaller circuits which can be executed within the constraints of available quantum hardware. The results of these smaller circuits are combined to reconstruct the outcome of the original problem. Circuit cutting can also be used to engineer gates between distant qubits which would otherwise require a large swap overhead.
+Circuit cutting is a technique to increase the size of circuits we can run on quantum hardware at the cost of an additional sampling overhead. A larger quantum circuit can be decomposed by cutting its gates or wires, resulting in smaller circuits that can be executed within the constraints of available quantum hardware. The results of these smaller circuits are combined to reconstruct the outcome of the original problem. Circuit cutting can also be used to engineer gates between distant qubits which would otherwise require a large swap overhead.
```
garrison: Being able to cut wires and gates is the same circuit is a key feature of this package.  Here is one way to express this explicitly, but I am open to others.

```suggestion
Circuit cutting is a technique to increase the size of circuits we can run on quantum hardware at the cost of an additional sampling overhead. A larger quantum circuit can be decomposed by cutting its gates, wires, or both, resulting in smaller circuits that can be executed within the constraints of available quantum hardware. The results of these smaller circuits are combined to reconstruct the outcome of the original problem. Circuit cutting can also be used to engineer gates between distant qubits which would otherwise require a large swap overhead.
```

docs/explanation/index.rst
```diff
@@ -6,29 +6,29 @@ Explanatory material
...
+There are `three settings <https://www.ibm.com/quantum/blog/circuit-knitting-with-classical-communication>`__ to consider for circuit cutting.  The first is where only local operations (LO) &mdash; as in, local *quantum* operations &mdash; are available.  The other settings introduce classical communication between the circuit executions, which is known in the quantum information literature as LOCC, for `local operations and classical communication <https://en.wikipedia.org/wiki/LOCC>`__.  The LOCC can be either near-time, one-directional communication between the circuit executions (the second setting), or real-time, bi-directional communication (the third setting).
```
garrison: will render by Sphinx as an mdash instead of literally as the text `&mdash;`

```suggestion
There are `three settings <https://www.ibm.com/quantum/blog/circuit-knitting-with-classical-communication>`__ to consider for circuit cutting.  The first is where only local operations (LO) --- as in, local *quantum* operations --- are available.  The other settings introduce classical communication between the circuit executions, which is known in the quantum information literature as LOCC, for `local operations and classical communication <https://en.wikipedia.org/wiki/LOCC>`__.  The LOCC can be either near-time, one-directional communication between the circuit executions (the second setting), or real-time, bi-directional communication (the third setting).
```

docs/explanation/index.rst
```diff
@@ -130,17 +130,17 @@ Wire cutting phrased as a two-qubit :class:`.Move` operation
...
 
-We have chosen to represent wire cuts in this way primarily because it is consistent with the way one must treat wire cuts when acting on physical qubits: for instance, a wire cut might take the state of physical qubit :math:`n` and continue it as physical qubit :math:`m` after the cut.  Our choice also has the benefit of allowing us to think of "instruction cutting" as a unified framework for considering both wire cuts and gate cuts in the same formalism, being that a wire cut is just a cut :class:`.Move` instruction.
+We have chosen to represent wire cuts in this way primarily because it is consistent with the required way to treat wire cuts when acting on physical qubits: for instance, a wire cut might take the state of physical qubit :math:`n` and continue it as physical qubit :math:`m` after the cut.  Our choice also has the benefit of allowing us to think of "instruction cutting" as a unified framework for considering both wire cuts and gate cuts in the same formalism, because a wire cut is just a cut :class:`.Move` instruction.
```
garrison:
```suggestion
We have chosen to represent wire cuts in this way primarily because it is consistent with the mental model for treating wire cuts when acting on physical qubits: for instance, a wire cut might take the state of physical qubit :math:`n` and continue it as physical qubit :math:`m` after the cut.  Our choice also has the benefit of allowing us to think of "instruction cutting" as a unified framework for considering both wire cuts and gate cuts in the same formalism, because a wire cut is just a cut :class:`.Move` instruction.
```

docs/explanation/index.rst
```diff
@@ -130,17 +130,17 @@ Wire cutting phrased as a two-qubit :class:`.Move` operation
...
+In this package, the number of samples taken from the distribution is generally controlled by a ``num_samples`` argument, and each sample has an associated weight that is used during expectation value reconstruction. Each weight with absolute value above a threshold of 1 / ``num_samples`` will be evaluated exactly.  The remaining low-probability elements &mdash; those in the tail of the distribution &mdash; will then be sampled, resulting in at most ``num_samples`` unique weights. Setting ``num_samples`` to infinity indicates that all weights should be generated rigorously, rather than by sampling from the distribution.
```
garrison:
```suggestion
In this package, the number of samples taken from the distribution is generally controlled by a ``num_samples`` argument, and each sample has an associated weight that is used during expectation value reconstruction. Each weight with absolute value above a threshold of 1 / ``num_samples`` will be evaluated exactly.  The remaining low-probability elements --- those in the tail of the distribution --- will then be sampled, resulting in at most ``num_samples`` unique weights. Setting ``num_samples`` to infinity indicates that all weights should be generated rigorously, rather than by sampling from the distribution.
```

INSTALL.rst
```diff
@@ -44,7 +42,7 @@ Note: If you are using Windows, use the following commands in PowerShell:
 Option 1: Install from PyPI
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-The most straightforward way to install the ``qiskit-addon-cutting`` package is via PyPI.
+The most straightforward way to install the ``qiskit-addon-cutting`` package is by using PyPI.
```
garrison: pip installs the package _from_ PyPI, so this is one potential better way to word this.

```suggestion
The most straightforward way to install the ``qiskit-addon-cutting`` package is by using pip.
```

docs/explanation/index.rst
abbycross: I think unicode will work but I'm not sure how translation will deal with the three hyphens (which is why I have been using `&mdash;` in the first place). Will we still be running sphinx once the docs have migrated?

abbycross:
```suggestion
Much of the circuit cutting literature describes a process where we sample from the distribution and take a single shot, then sample from the distribution again and repeat; however, this is not feasible in practice, so we instead perform all sampling upfront.  For now, because of limitations in version 1 of the Qiskit primitives, we take a fixed number of shots for each considered subexperiment and send the subexperiments to the backend(s) in batches. During reconstruction, each subexperiment contributes to the final result with proportion equal to its weight.  We must ensure the number of shots taken is sufficient for the heaviest weighted subexperiment.  An `open issue (#532) <https://github.com/Qiskit/qiskit-addon-cutting/issues/532>`__ tracks work to support passing an individual ``shots`` count with each subexperiment to Qiskit Runtime, so that each subexperiment will be run with a number of shots proportional to that subexperiment's weight in the final result.  This per-experiment shots count is a new feature enabled by version 2 of the Qiskit primitives.
```

abbycross: @garrison how is this rewording?

garrison: @eharvey328 can clarify -- but my understanding is that the documentation pipeline will run Sphinx and use its output as input to a later stage of the build.  Assuming this is correct, then `---` will be converted to the mdash unicode character when Sphinx runs.

garrison: It's great! Thank you

abbycross: Ah I see! That would indeed make things easier. @eharvey328 would `--` also get converted to emdash in unicode?

garrison: I believe that Sphinx will convert `--` to an ndash instead.

INSTALL.rst
```diff
@@ -54,27 +52,27 @@ The most straightforward way to install the ``qiskit-addon-cutting`` package is
...
+If you plan to develop in the repository or run the notebooks locally, install from source.
 
 In either case, the first step is to clone the ``qiskit-addon-cutting`` repository.
 
 .. code:: sh
 
     git clone git@github.com:Qiskit/qiskit-addon-cutting.git
     
-Next, upgrade pip and enter the repository. 
+Next, upgrade `pip` and enter the repository.
```
garrison:
```suggestion
Next, upgrade ``pip`` and enter the repository. 
```

garrison (approved):

PR #829, opened 2026-09-16
Adopt client-side (Executor-backed) SamplerV2 in docs and tests

garrison:
This migrates the documentation notebooks and the tests that use `SamplerV2` from the **server-side** implementation (imported from `qiskit_ibm_runtime`) to the new **client-side**, Executor-backed implementation in `qiskit_ibm_runtime.executor_sampler`, introduced in qiskit-ibm-runtime 0.49.0.

## What changed

- **Imports.** `from qiskit_ibm_runtime import SamplerV2` → `from qiskit_ibm_runtime.executor_sampler import SamplerV2`. Where a line imported both `SamplerV2` and `Batch`, it is split so that `Batch` stays on the top-level import.
  - `docs/tutorials/01_gate_cutting_to_reduce_circuit_width.ipynb`
  - `docs/tutorials/02_gate_cutting_to_reduce_circuit_depth.ipynb`
  - `docs/tutorials/03_wire_cutting_via_move_instruction.ipynb`
  - `docs/how-tos/how_to_provide_multiple_observables.ipynb`
  - `test/test_cutting_roundtrip.py`
  - `test/test_cutting_workflows.py`
- **Version floor.** Bumped `qiskit-ibm-runtime` to `>=0.49.0` in the `test` and `notebook-dependencies` extras, since the `executor_sampler` module does not exist before that release. This is a test/docs-only extra, not a core dependency of the library, so it does not affect end users.

## Why

The interface, options tree, `run()` signature, PUB format, and result structure are unchanged between the server-side and client-side V2 primitives, so the migration is limited to the import. The client-side implementation captures all design intent on the client and runs it through a single server-side Executor primitive, giving a white-box execution path.

Note that the explicit `executor_sampler` import is required today; in a future release the top-level `from qiskit_ibm_runtime import SamplerV2` will resolve to the client-side implementation directly.

## Testing

`test_cutting_roundtrip.py` and `test_cutting_workflows.py` construct `SamplerV2(AerSimulator())` and run it in local mode, so they exercise the client-side path end to end. Both files pass (44 tests). The `Options ... have no effect in local testing mode` warning is the expected benign notice that error-mitigation options are inert against a simulator.

<hr>

This PR was generated by Claude Opus 4.8 (1M context) under my guidance.


garrison:
Marking as a draft until I gain clarity on whether a future version of qiskit-ibm-runtime will make the top-level Sampler be the executor-based one.

